PR #63, opened 2025-09-12
Add Div primary operation

LindaGuiga:
This PR aims at adding `Div` using the `Mul` AIR.


sai-deng:
Sorry, my previous commits are not signed. I need to re-sign all of them, which means you may need to drop this branch and apply the diff to a new fork from main.

PR #82, opened 2025-09-18
Implement `Recursive` for `FriProofTargets`

LindaGuiga:
This PR aims at implementing `Recursive` for `FriProofTargets`, which is the recursive version of `FriProof`. It also implements `RecursiveMmcs`/`RecursiveExtensionMmcs` for Mmcs where the commitments are hashes. 
This is the first step to having `TwoAdicFriPcs` implement `RecursivePcs`, since its `Proof` is `FriProof`

I added a test creating new ProofTargets and setting them, for a Fibonacci proof. For this reason, to avoid code duplication for `FibonacciAir`, I added a `test_utils.rs` in the `circuit` folder containing the definition of FibonacciAir. 

The test is called `test_fibonacci_verifier` as I wanted this to use `verify_circuit` once all the pieces are put together. To that avail, I also added an implementation of `RecursiveAir` for `FibonacciAir` in a `test_utils.rs`.

Note: for now, base field elements are turned into extension field ones. This can be updated later with packing.


sai-deng (approved):
Thanks.

Issue #73, opened 2025-09-15
Add a MerkleVerify gate


4l0n50:
Implement a gate that takes `leaf`, `index`, and `root` as public inputs and verifies a Merkle path from leaf to root, with the path provided as a private witness. Also add trace generation for [MerkleTreeAir](https://github.com/Plonky3/Plonky3-recursion/blob/8289e045b105077db0a441e0aab9d0e7bbb41330/merkle-tree-air/src/air.rs#L34)

PR #85, opened 2025-09-19
Add merkle verify gate

4l0n50:
Closes https://github.com/Plonky3/Plonky3-recursion/issues/73.

### DIFF
--- a/circuit-prover/src/config.rs
+++ b/circuit-prover/src/config.rs
@@ -121,7 +121,6 @@ pub mod babybear_config {
         build_standard_config_generic::<BB, _, 4>(perm)
     }
 
-    // TODO: For now we are not considering packed inputs for BabyBear
     pub fn baby_bear_standard_compression_function() -> impl PseudoCompressionFunction<[BB; 8], 2> {
         let perm = default_babybear_poseidon2_16();
         standard_compression_function_generic::<BB, _>(perm)
@@ -142,7 +141,6 @@ pub mod koalabear_config {
         build_standard_config_generic::<KB, _, 4>(perm)
     }
 
-    // TODO:For now we are not considering packed inputs for KoalaBear
     pub fn koala_bear_standard_compression_function() -> impl PseudoCompressionFunction<[KB; 8], 2>
     {
         let perm = default_koalabear_poseidon2_16();
@@ -165,7 +163,6 @@ pub mod goldilocks_config {
         build_standard_config_generic::<GL, _, 2>(perm)
     }
 
-    // TODO: For now we are considering packed inputs for Goldilocks
     pub fn goldilocks_standard_compression_function() -> impl PseudoCompressionFunction<[GL; 8], 2>
     {
         let mut rng = SmallRng::seed_from_u64(1);


circuit-prover/src/prover.rs
```diff
@@ -282,12 +283,15 @@ where
...
         let mul_proof = prove(&self.config, &mul_air, mul_matrix, pis);
 
-        // FakeMerkle (always uses base field regardless of traces D)
-        let fake_merkle_matrix =
-            FakeMerkleVerifyAir::<F>::trace_to_matrix(&traces.fake_merkle_trace);
-        let fake_merkle_air =
-            FakeMerkleVerifyAir::<F>::new(traces.fake_merkle_trace.left_values.len());
-        let fake_merkle_proof = prove(&self.config, &fake_merkle_air, fake_merkle_matrix, pis);
+        let merkle_matrix = match D {
+            1..=8 => Ok(MerkleVerifyAir::<F, 8, 32>::trace_to_matrix(
```
LindaGuiga, line 287: Don't we want to have `DIGEST_ELEMS` be generic?

circuit/src/tables.rs
```diff
@@ -178,21 +189,173 @@ pub struct MulTrace<F> {
 
 /// Fake Merkle verification table (simplified: single field elements)
 #[derive(Debug, Clone)]
-pub struct FakeMerkleTrace<F> {
+pub struct MerkleTrace<F> {
+    /// All the merkle paths computed in this trace
+    pub merkle_paths: Vec<MerklePathTrace<F>>,
+}
+
+/// A single Merkle Path verification table (simplified: single field elements)
```
LindaGuiga, line 197:
```suggestion
/// A single Merkle Path verification table
```

circuit/src/tables.rs
```diff
@@ -178,21 +189,173 @@ pub struct MulTrace<F> {
...
+#[derive(Debug, Clone, Default)]
+pub struct MerklePathTrace<F> {
     /// Left operand values (current hash)
-    pub left_values: Vec<F>,
+    pub left_values: Vec<Vec<F>>,
```
LindaGuiga, line 201: Maybe add a comment to explain why each value is a Vec? Also, is there a way to turn the vec into an array (like [F; DIGEST_ELEMS])?

circuit/src/tables.rs
```diff
@@ -178,21 +189,173 @@ pub struct MulTrace<F> {
...
+            // TODO: What is the address of this value?
+            trace.left_index.push(leaf_index); // Points to witness bus
+
+            // Sibling becomes right operand (private data - not on witness bus)
+            trace.right_values.push(
+                Self::into_f_slice::<_, DIGEST_ELEMS, BF_DIGEST_ELEMS>(&sibling_value)?.to_vec(),
+            );
+            trace.right_index.push(0); // Not on witness bus - private data
+
+            // Compute parent hash (simple mock hash: left + right + direction)
```
LindaGuiga, line 322:
```suggestion
            // Compute parent hash 
```

circuit/src/config.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    pub fn default_babybear_poseidon2_circuit_runner_config() -> DefaultBabyBearConfig {
+        let permutation = default_babybear_poseidon2_16();
+        let compress = TruncatedPermutation::<_, 2, 8, 16>::new(permutation);
+        DefaultBabyBearConfig { compress }
+    }
+}
+
+pub mod koalabear_config {
```
LindaGuiga, line 37: I think at least some of these configs are in circuit-prover, already, right? If so, is there a way to unify the two? (as in put them in whichever folder that will not cause cyclic imports)

circuit/src/config.rs
```diff
@@ -0,0 +1,93 @@
+use p3_field::PrimeCharacteristicRing;
+use p3_symmetric::PseudoCompressionFunction;
+
+pub trait CircuitRunnerConfig<const DIGEST_ELEMS: usize> {
```
sai-deng, line 4: I don’t think this trait is necessary. Same for #86 , we can reuse the ADD/MUL table pattern to handle the varying column counts issue.

merkle-tree-air/src/cols.rs
```diff
@@ -0,0 +1,201 @@
...
+use itertools::izip;
+use p3_circuit::tables::MerkleTrace;
+use p3_field::{BasedVectorSpace, Field};
+use p3_matrix::dense::RowMajorMatrix;
+
+use crate::air::MerkleVerifyAir;
+
+#[derive(Debug)]
+#[repr(C)]
+pub struct MerkleTreeCols<T, const DIGEST_ELEMS: usize, const MAX_TREE_HEIGHT: usize> {
```
sai-deng, line 12: We would not need it if we follow the pattern in #81

circuit/src/config.rs
4l0n50: Removed

circuit-prover/src/config.rs
```diff
@@ -76,14 +81,17 @@ pub type ProverConfig<F, P, const CD: usize> = StarkConfig<
...
 where
     F: StarkField + BinomiallyExtendable<CD>,
-    P: StarkPermutation<F>,
+    P: StarkPermutation<F> + Clone + 'static,
+    EF: BasedVectorSpace<F>,
 {
     let hash = MyHash::<P, 16, 8, 8>::new(perm.clone());
     let compress = MyCompress::<P, 2, 8, 16>::new(perm.clone());
-    let val_mmcs = ValMmcs::<F, P>::new(hash, compress);
+    let val_mmcs = ValMmcs::<F, P>::new(hash, compress.clone());
```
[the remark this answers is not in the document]
Nashtare: Yeah that would suffice (but the surface area that needs fixing is bigger, the constants are spread a bit everywhere.
This can be left for a future PR if it's too much

circuit-prover/src/config.rs
```diff
@@ -76,14 +81,17 @@ pub type ProverConfig<F, P, const CD: usize> = StarkConfig<
 /// - Two-adic FRI PCS for polynomial commitments
 /// - Merkle tree MMCS for vector commitments  
 /// - Duplex challenger for Fiat-Shamir
-pub fn build_standard_config_generic<F, P, const CD: usize>(perm: P) -> ProverConfig<F, P, CD>
+pub fn build_standard_config_generic<EF, F, P, const CD: usize>(
+    perm: P,
+) -> (ProverConfig<F, P, CD>, MerkleVerifyConfig<EF>)
```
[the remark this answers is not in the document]
4l0n50: I tried to do that, but then some places where assuming that ProverConfig was just an alias for StarkConfig. It wouldn't be big deal to retrieve the starkconfig, but I decided to return the tuple here because it required less changes in the code. Btw, I think it's still not clear where these configs should live

merkle-tree-air/src/air.rs
```diff
@@ -1,186 +1,202 @@
...
-        for i in 0..DIGEST_ELEMS {
-            for j in 0..DIGEST_ELEMS {
-                cur_to_hash[i] += local.height_encoding[j].clone()
-                    * (local.index_bits[j].clone() * local.sibling[j].clone()
-                        + (AB::Expr::ONE - local.index_bits[j].clone()) * local.state[j].clone());
-                cur_to_hash[DIGEST_ELEMS + i] += local.index_bits[j].clone()
-                    * (local.index_bits[j].clone() * local.sibling[j].clone()
-                        + (AB::Expr::ONE - local.height_encoding[j].clone())
-                            * local.state[j].clone());
+        let mut cur_to_hash = vec![AB::Expr::ZERO; 2 * self.config.digest_elems];
```
sai-deng, line 184: I think we should move this logic to a CTL.

LindaGuiga: That is the goal, but since we don't have CTLs yet, this was to define the columns that we are sending (CTLs take expressions for the values and the filters anyway). Otherwise, `cur_to_hash` is unused for now...

merkle-tree-air/src/air.rs
```diff
@@ -1,186 +1,202 @@
...
             main.row_slice(1).expect("The matrix only has 1 row?"),
         );
-        let local: &MerkleVerifyCols<AB::Var, DIGEST_ELEMS, MAX_TREE_HEIGHT> = (*local).borrow();
-        let next: &MerkleVerifyCols<AB::Var, DIGEST_ELEMS, MAX_TREE_HEIGHT> = (*next).borrow();
+
+        let index_bits = &local[self.index_bits()];
+        let length = &local[self.length()];
+        let next_length = &next[self.length()];
+        let sibling = &local[self.sibling()];
+        let state = &local[self.sibling()];
```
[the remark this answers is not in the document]
4l0n50: ops!

merkle-tree-air/src/air.rs
Nashtare: We know the system will be unsound until we have a proper lookup argument across chips, so I think it would be fine to just explicitly write that we expect to have the CTL logic handle this part in the future? anyway non blocking

merkle-tree-air/src/air.rs
```diff
@@ -1,186 +1,202 @@
...
-///
-/// `MAX_TREE_HEIGHT` is the maximal tree height that can be handled by the AIR.
-pub struct MerkleVerifyAir<F, C, const DIGEST_ELEMS: usize, const MAX_TREE_HEIGHT: usize>
-where
-    F: Field,
-{
-    pub compress: C,
-    _phantom: PhantomData<F>,
+
+/// Configuration for the merkle table AIR rows.
```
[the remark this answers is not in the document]
4l0n50: Is it ok now?

sai-deng:
> As discussed with @hratoanina and @LindaGuiga — we’re not supporting packing yet. So I added a, for now disabled, packing option to Merkle gates.

Could you please explain why packing is disabled for now?

circuit-prover/src/air/merkle_verify_air.rs
```diff
@@ -0,0 +1,146 @@
...
+        type MyCompress = CompressionFunctionFromHasher<U64Hash, 2, 4>;
+        let compress = MyCompress::new(u64_hash);
+
+        const NUM_INPUTS: usize = 4;
+        const HEIGHT: usize = 8;
+        const DIGEST_ELEMS: usize = 8;
+
+        // Generate random inputs.
+
+        let mut rng = SmallRng::seed_from_u64(1);
```
sai-deng, line 61: Why using random inputs here? (You already compute real values in the example.)

circuit-prover/src/air/merkle_verify_air.rs
```diff
@@ -0,0 +1,146 @@
...
+    use p3_circuit::WitnessId;
+    use p3_circuit::tables::{MerklePrivateData, MerkleTrace};
+    use p3_keccak::{Keccak256Hash, KeccakF};
+    use p3_merkle_tree_air::air::MerkleVerifyAir;
+    use p3_symmetric::{CompressionFunctionFromHasher, PaddingFreeSponge, SerializingHasher};
+
+    use crate::config::babybear_config::build_standard_config_babybear;
+
+    #[test]
+    fn prove_mmcs_verify_poseidon() -> Result<
```
sai-deng, line 19: The purpose of this test is unclear to me. Should we move it into the merkle-tree-air crate?

circuit-prover/src/config.rs
```diff
@@ -76,14 +81,18 @@ pub type ProverConfig<F, P, const CD: usize> = StarkConfig<
 /// - Two-adic FRI PCS for polynomial commitments
 /// - Merkle tree MMCS for vector commitments  
 /// - Duplex challenger for Fiat-Shamir
-pub fn build_standard_config_generic<F, P, const CD: usize>(perm: P) -> ProverConfig<F, P, CD>
+pub fn build_standard_config_generic<EF, F, P, const CD: usize>(
```
sai-deng, line 84: I think we should not make a merkle_config here. We can add a new function to make a merkle config.

circuit-prover/src/config.rs
```diff
@@ -104,11 +164,15 @@ pub mod babybear_config {
 
     use super::*;
 
-    pub type BabyBearConfig = ProverConfig<BB, Poseidon2BB<16>, 4>;
+    pub type BabyBearConfig<F> = (ProverConfig<BB, Poseidon2BB<16>, 4>, MerkleVerifyConfig<F>);
```
sai-deng, line 167: Since MerkleVerify is an optional Op in the circuit/prover, we should not have it in the Config type.

LindaGuiga:
> > As discussed with @hratoanina and @LindaGuiga — we’re not supporting packing yet. So I added a, for now disabled, packing option to Merkle gates.
> 
> Could you please explain why packing is disabled for now?

The problem with fully packing everything (so for example, when creating a proof structure), is that the challenger sometimes needs to observe both base field and extension field elements, without adding any padding in between. So for example, if you have to observe `degree_bits` (one base field element, and we do it at the start of the verifier) then elements that are packed, you might overflow the sponge rate (since, because of that first base field element, we would have something that is no longer a multiple of the rate). The correct behaviour then would be to fill the rate with base field elements (therefore breaking up packed elements), apply the permutation, and then add the rest of the base field elements. But for now, given the current Witness table, we don't have a way to select one base field element among the D in a packed extension field element. We can change the Witness table to support the selection of individual elements, though.

sai-deng:
> > > As discussed with @hratoanina and @LindaGuiga — we’re not supporting packing yet. So I added a, for now disabled, packing option to Merkle gates.
> 
> > 
> 
> > Could you please explain why packing is disabled for now?
> 
> 
> 
> The problem with fully packing everything (so for example, when creating a proof structure), is that the challenger sometimes needs to observe both base field and extension field elements, without adding any padding in between. So for example, if you have to observe `degree_bits` (one base field element, and we do it at the start of the verifier) then elements that are packed, you might overflow the sponge rate (since, because of that first base field element, we would have something that is no longer a multiple of the rate). The correct behaviour then would be to fill the rate with base field elements (therefore breaking up packed elements), apply the permutation, and then add the rest of the base field elements. But for now, given the current Witness table, we don't have a way to select one base field element among the D in a packed extension field element. We can change the Witness table to support the selection of individual elements, though.

I think the correct and most efficient way to solve it is to change Plonky3 prover to only observe extension field elements. We can add a TODO for now.

circuit-prover/src/config.rs
4l0n50: no longer there

circuit-prover/src/config.rs
4l0n50: no longer there

merkle-tree-air/src/air.rs
```diff
@@ -1,186 +1,221 @@
 use alloc::vec;
```
sai-deng, line 1: I think we should rename all “merkle verify” to “mmcs verify,” since the former doesn’t involve extra folding and could be misleading for other use cases that need a true Merkle verification.

circuit-prover/examples/merkle_verify.rs
```diff
@@ -0,0 +1,143 @@
...
+        leaf.to_vec(),
+        |state, (direction, (sibling, other_sibling))| {
+            let (left, right) = if *direction {
+                (state.clone(), sibling.clone())
+            } else {
+                (sibling.clone(), state.clone())
+            };
+            let mut new_state = (merkle_config.compress)([&left, &right]);
+            if let Some(other_sibling) = other_sibling {
+                new_state = (merkle_config.compress)([&state, other_sibling]);
```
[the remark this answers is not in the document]
4l0n50: Yes, and I added a test [here](https://github.com/Plonky3/Plonky3-recursion/commit/8295a0c9cb56420e163ea10b34d546bccfcac0ad)

merkle-tree-air/src/air.rs
sai-deng: I would remove it, since the expression is wrong (it adds some old values twice) and it doesn’t handle packing.

sai-deng (approved):
@4l0n50 I pushed a rename commit and another commit with several fixes. Please review and merge the PR as soon as possible to unblock others.

My fixes include:
Adding root indices to the AIR tables.
Fixing missing or incorrect constraints in AIR tables.
Correcting the logic for hashing extra siblings.

4l0n50 (commented):
I'm ok with your changes. Thanks for you help!

PR #86, opened 2025-09-19
WIP: Add sponge ops

hratoanina:
Opening as draft for visibility and for the API. I'm having trouble making it generic for the hash function. It technically compiles and runs, but it's not very generic for now.

### DIFF
--- a/circuit/src/tables.rs
+++ b/circuit/src/tables.rs
@@ -1,5 +1,11 @@
-use p3_field::PrimeCharacteristicRing;
+use core::array;
 
+use p3_field::{Field, PrimeCharacteristicRing};
+use p3_symmetric::{CryptographicHasher, CryptographicPermutation};
+use rand::SeedableRng;
+use rand::rngs::SmallRng;
+
+use crate::builder::{CIRCUIT_HASH_CAPACITY, CIRCUIT_HASH_RATE, CircuitPerm};
 use crate::circuit::Circuit;
 use crate::op::{NonPrimitiveOpPrivateData, Prim};
 use crate::types::{NonPrimitiveOpId, WitnessId};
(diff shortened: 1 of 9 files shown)


circuit/src/tables.rs
```diff
@@ -222,6 +258,35 @@ impl<
...
+            add_trace,
+            mul_trace,
+            sub_trace,
+            fake_merkle_trace,
+            sponge_trace: SpongeTrace::default(),
+        })
+    }
+
+    /// Run the circuit and generate traces
+    pub fn run_with_hash<P: CryptographicPermutation<[F; N]>, const N: usize>(
```
hratoanina, line 274: Ugly, but I don't want to add the generics and bounds to the normal one.

More generally, it feels like it would be better to have a single `run` function handling all possible cases without all the generics. The only other solution is to somehow put all future generics and trait bounds in the CircuitBuilder/Circuit/CircuitRunner structs themselves, but it doesn't feel ideal either. Not sure how to do it properly.

Nashtare (commented):
added my two cents, but not too deep yet given this is still WIP

hratoanina:
I merged with the current design, which is veri different from Alonso's approach with a dedicated config per non primitive op. If Alonso's approach is favoured it should be merged before this PR, and I'll merge on top of it.

circuit/src/tables.rs
```diff
@@ -398,58 +454,64 @@ impl<F: CircuitField> CircuitRunner<F> {
...
-            };
-
-            // Clone private data option to avoid holding a borrow on self
-            if let Some(Some(NonPrimitiveOpPrivateData::FakeMerkleVerify(private_data))) =
-                self.non_primitive_op_private_data.get(op_idx).cloned()
-            {
-                let mut current_hash =
-                    if let Some(val) = self.witness.get(leaf.0 as usize).and_then(|x| x.as_ref()) {
-                        *val
+            // Only handle FakeMerkleVerify ops here
```
4l0n50, line 457: Why only FakeMerkle here?

circuit/src/tables.rs
```diff
@@ -174,14 +214,28 @@ impl<F: CircuitField> CircuitRunner<F> {
             ) => {
                 // Type match - good!
             }
+            (crate::op::NonPrimitiveOp::HashAbsorb { .. }, _) => {
+                panic!("HashAbsorb operation does not take private data");
```
4l0n50, line 218: you could add the result of the operation in the private data. In this way you can remove the need of passing the permutation to the runner.

PR #79, opened 2025-09-17
Add verify_circuit

LindaGuiga:
This PR aims at adding `verify_circuit`, which is the recursive equivalent of verify.
It relies on the traits defined in recursive_traits.rs.

closes #31

PR #87, opened 2025-09-19
Implement RecursiveAir for Air

LindaGuiga:
Small PR to implement `RecursiveAir` for `Air`. It can make #82, #79  and future PRs slightly easier.


Nashtare (approved):
LGTM

Issue #58, opened 2025-09-10
Implement a circuit version of `sample_bits` for the Challenger


LindaGuiga:
The challenger is used to get an `index` with `n` bits in FRI verification. This index is determined by using `sample` on the challenger and then XORing with `((1 << bits) - 1)`. Thus, we need to be able to simulate that in the recursive circuit.

PR #93, opened 2025-09-23
circuit: add a circuit version of `sample_bits` for the Challenger

tcoratger:
Should close https://github.com/Plonky3/Plonky3-recursion/issues/58

### DIFF
--- a/recursion/src/circuit_fri_verifier.rs
+++ b/recursion/src/circuit_fri_verifier.rs
@@ -96,7 +96,6 @@ pub fn verify_query<F: Field>(
     builder.connect(folded_eval, final_value);
 }
 
-
 /// Compute x₀ for phase `i` from the query index bits and a caller-provided power ladder.
 ///
 /// For phase with folded height `k` (log_folded_height), caller must pass:
--- a/circuit/src/utils.rs
+++ b/circuit/src/utils.rs
@@ -375,8 +375,8 @@ mod tests {
 
         // Verify the bits are correctly decomposed - 6 = [0,1,1] in little-endian
         assert_eq!(traces.public_trace.values[0], BabyBear::ZERO); // bit 0
-        assert_eq!(traces.public_trace.values[1], BabyBear::ONE);  // bit 1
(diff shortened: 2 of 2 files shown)


sai-deng:
Should this be an AIR table? How many rows are expected in the verification circuit?

tcoratger:
@sai-deng Ok I thought about that and I think this probably deserves a bit more thinking due to the fact that the number of queries could depend on different things such as the targeted number of bits of security as well as the chosen conjecture for the proximity gap thing for Reed Solomon code. As WHIR is a bit different in this perspective.

So that to move forward and check the small details of this later, I've proposed here a different approach with just a scaffolding of what needs to be done with the important objects and functions in place that can be filled later.

Let me know if this approach sounds good to you.

sai-deng:
> @sai-deng Ok I've made a much simpler approach in the circuit context.
> 
> We can iterate over there and go back to the AIR version if needed in the future.
> 
> Sorry for the back and forth.

Thanks!

I think we can also revert the changes in table.rs?

#108 is related. We should be able to automatically set these witness slots.

tcoratger:
@sai-deng fixed

recursion/src/circuit_fri_verifier.rs
```diff
@@ -113,6 +113,31 @@ pub fn reconstruct_index_from_bits<F: Field>(
...
+///     x = Σ b_i · 2^i
+pub fn decompose_to_bits<F: Field, const N_BITS: usize>(
+    builder: &mut CircuitBuilder<F>,
+    x: Target,
+) -> Vec<Target> {
+    let mut bits = Vec::with_capacity(N_BITS);
+
+    // Create bit witness variables
+    for _ in 0..N_BITS {
+        let bit = builder.add_public_input(); // TODO: Should be witness
```
sai-deng, line 129: How about moving this and reconstruct_index_from_bits() to utils.rs (reused by the challenger). Please also include a unit test.

sai-deng (approved):
Thanks!

PR #94, opened 2025-09-23
chore: use `thiserror` crate for error handling

tcoratger:


Nashtare:
I was also leaning towards `thiserror` as a personal preference. It's not too late to revisit our decision though, @sai-deng any particular objections against Thomas' view?

Nashtare (approved):

Issue #98, opened 2025-09-24
Modularize circuit builder: opt‑in non‑primitive ops


sai-deng:
We should make the circuit builder more modular so users can choose which non‑primitive ops to include in their circuits. This may involve:

  - Exposing non‑primitive ops via per‑op extension traits, with an optional allowlist/profile for runtime policy.
  - Adding a FRI‑focused alias/newtype.

PR #101, opened 2025-09-24
circuit builder: add some tests

tcoratger:
I've added some tests to discover the behaviour here. If this is interesting to you I can think about a smoother API for the circuit builder in a followup PR.


sai-deng (approved):
LGTM!

tcoratger:
> > If this is interesting to you I can think about a smoother API for the circuit builder in a followup PR.
> 
> A related Issue #98 and PR #102 by Robin modularizes the circuit builder. With his work, users can choose which non-primitive ops to include. Primitive ops are available in all builders by default, while non-primitive ops are opt-in. This enables e.g. the WHIR and FRI verifier builders to select different sets of non-primitive ops.

Amazing, will check this

PR #102, opened 2025-09-24
feat: add basic policy for non-primitive op handling at runtime

Nashtare:
Attempt at dealing with #98, although this may look fairly limited as of now given that `main` does not contain yet any non-primitive op apart from the fake merkle verify one, but opening it as draft nonetheless for visibility.

There are currently 2 profiles:

- `DefaultProfile` with no non-primitive ops allowed
- `AllowAllProfile` with all ops supported

@sai-deng does it look fine to you or were you envisioning a different approach for runtime selection? happy to tweak things around


sai-deng:
LGTM! I’m also thinking about Alonso’s PR #85 — can we scope the op‑specific generics/config to the op itself (and its runtime executor) rather than the builder? That way the builder stays generic‑free, and only the op’s extension/runner path carries the generics.

PR #109, opened 2025-09-25
[wip] Finish fri verifier circuits

sai-deng:


4l0n50 (approved):
I just missed more links with the non recursive counterpart and more modularization in the test (maybe this carries from an older PR?), but nothing serious. So, LGTM

recursion/src/circuit_fri_verifier.rs
```diff
@@ -178,3 +183,311 @@ pub fn verify_query_from_index_bits<F: Field>(
...
+/// - Add recursive MMCS verification for both input openings (open_input) and per-phase commitments.
+#[allow(clippy::too_many_arguments)]
+pub fn verify_fri_circuit<F, EF, RecMmcs, Inner, Witness>(
+    builder: &mut CircuitBuilder<EF>,
+    fri_proof_targets: &FriProofTargets<F, EF, RecMmcs, InputProofTargets<F, EF, Inner>, Witness>,
+    alpha: Target,
+    betas: &[Target],
+    index_bits_per_query: &[Vec<Target>],
+    challenge_points: &[Target],
+    challenge_point_values: &[Vec<Target>],
```
LindaGuiga, line 352: Why not use `CommsWithOpeningPoints`, which would be more similar to the original function, reduce the number of arguments, and then use the commitments for Mmcs checks later on?

This would also be closer to the signature in the `RecursivePcs` trait that we currently have

recursion/src/circuit_fri_verifier.rs
```diff
@@ -178,3 +183,311 @@ pub fn verify_query_from_index_bits<F: Field>(
...
+        let mut roll_ins: Vec<Option<Target>> = vec![None; num_phases];
+        for &(h, ro) in reduced_by_height.iter().skip(1) {
+            let i = log_max_height
+                .checked_sub(1)
+                .and_then(|x| x.checked_sub(h))
+                .expect("height->phase mapping underflow");
+            if i < num_phases {
+                roll_ins[i] = Some(match roll_ins[i] {
+                    None => ro,
+                    Some(prev) => builder.add(prev, ro),
```
LindaGuiga, line 473: Can this happen? Looking at the original code, they only check the next iterator (`ro_iter`) value once, so shouldn't this case return an error instead?

recursion/src/circuit_fri_verifier.rs
```diff
@@ -178,3 +183,311 @@ pub fn verify_query_from_index_bits<F: Field>(
...
+        builder.assert_bool(b);
+    }
+    debug_assert_eq!(
+        index_bits.len(),
+        log_max_height,
+        "index_bits.len() must equal log_max_height"
+    );
+
+    // height -> (alpha_pow_for_this_height, ro_sum_for_this_height)
+    let mut by_height: BTreeMap<usize, (Target, Target)> = BTreeMap::new();
```
LindaGuiga, line 293: If I understand correctly, this function corresponds to `open_input`. But in that case, are we not implementing batch fri? (if I'm not mistaken, the batching case would have an additional loop for each batch commitment and opening proof)

tcoratger (approved):
Looks good to me, just minor comments. I noticed that in several functions you used `EF` as a generic despite the fact that there's no `F` (in this case `EF` would be the extension of `F`). I assume this is to indicate that it's the extension field of the verifier, and that it's obvious to the reader. But if there's only one field defined, by convention in Plonky3, aren't we used to simply calling it `F`, whether it's actually an extension field or not?

Just a suggestion; both are fine to me :)

recursion/src/circuit_fri_verifier.rs
```diff
@@ -178,3 +183,311 @@ pub fn verify_query_from_index_bits<F: Field>(
...
+        .collect();
+    v.sort_by(|a, b| b.0.cmp(&a.0)); // descending by height
+    v
+}
+
+/// Verify FRI arithmetic in-circuit.
+///
+/// TODO:
+/// - Introduce `RecursiveChallenger` and derive `alpha`, `betas`, and `index_bits_per_query`
+///   internally (observe commitments, sample challenges, check PoW, observe final poly coeffs).
```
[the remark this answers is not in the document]
sai-deng: Updated the comments.

recursion/src/circuit_fri_verifier.rs
sai-deng: Thanks for the catch! Yes, multi rounds input opening is not yet supported. I’ll address it in the next PR. Added a TODO.

recursion/src/circuit_fri_verifier.rs
sai-deng: Added a TODO.

recursion/src/circuit_fri_verifier.rs
```diff
@@ -178,3 +183,311 @@ pub fn verify_query_from_index_bits<F: Field>(
...
+        *alpha_pow_h = new_alpha_pow_h;
+    }
+
+    // Into descending (height, ro) list
+    let mut v: Vec<(usize, Target)> = by_height
+        .into_iter()
+        .map(|(h, (_ap, ro))| (h, ro))
+        .collect();
+    v.sort_by(|a, b| b.0.cmp(&a.0)); // descending by height
+    v
```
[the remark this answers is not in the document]
sai-deng: Thanks for the catch!

sai-deng:
> Looks good to me, just minor comments. I noticed that in several functions you used `EF` as a generic despite the fact that there's no `F` (in this case `EF` would be the extension of `F`). I assume this is to indicate that it's the extension field of the verifier, and that it's obvious to the reader. But if there's only one field defined, by convention in Plonky3, aren't we used to simply calling it `F`, whether it's actually an extension field or not?
> 
> Just a suggestion; both are fine to me :)

Thanks for the review! For recursion circuits I think it is clearer to the current pattern: EF for the circuit field in single‑field helpers, and F + EF when both base and circuit fields appear.

PR #114, opened 2025-09-29
circuit: small reorg

tcoratger:
- To limit the number of lines in the tables.rs file that is quite heavy, this is probably better to have a separate dedicated file.
- I've seen 2 usages (and probably more in the future) of similar trait bounds for the field in circuit, we could define an abstraction for this.

### DIFF
--- a/circuit/src/tables.rs
+++ b/circuit/src/tables.rs
@@ -1,86 +1,10 @@
-use alloc::string::String;
 use alloc::vec::Vec;
 use alloc::{format, vec};
-
-use p3_field::Field;
-use thiserror::Error;
 
 use crate::circuit::Circuit;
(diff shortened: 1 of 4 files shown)


circuit/src/tables.rs
```diff
@@ -517,13 +431,8 @@ impl<
...
                     // Compute parent hash (simple mock hash: left + right + direction)
-                    let parent_hash = current_hash
-                        + *sibling_value
-                        + if direction {
-                            F::from_u64(1)
-                        } else {
-                            F::from_u64(0)
-                        };
+                    let parent_hash =
+                        current_hash + *sibling_value + if direction { F::ONE } else { F::ZERO };
```
tcoratger, line 435: This way is faster

PR #115, opened 2025-09-30
Add support for multi rounds input folding in the FRI verification circuit

sai-deng:


sai-deng:
> Thank you!

Thank you for the quick reviews!

PR #118, opened 2025-10-01
[WIP] Design for recursive lookups (global and local interactions)

LindaGuiga:
This is still work in progress, but I am opening a draft PR so that we can decide early on whether we can agree on this design.

The goal is to create a design for recursive lookups.
The trait `RecursiveAirLookupVerification` contains the methods necessary for an AIR to check its own lookup constraints. Thus, it enables a `RecursiveAir` to:
- get the permutation columns (the auxiliary columns, like the running sum)
- get the challenges associated to the local lookups
- for local lookups, check that the auxiliary columns are updated correctly, and that the final cumulative sum (for LogUp) is 0
- for global lookups, check that the auxiliary columns are updated correctly, and that the final cumulative sum (for LogUp) is the same as the one provided by the prover.

Note that for global lookups (= global interactions), we need the index of the table with which we are interacting, as well as the direction of the lookup (sending or receiving), to determine whether the multiplicities should be negated. This is why I introduced `Direction` and `Table`.

In order to facilitate the use of airs without lookups, I introduced a wrapper `AirWithoutLookups` which implements `RecursiveAirLookupVerification`.

### DIFF
--- a/recursion/src/circuit_verifier.rs
+++ b/recursion/src/circuit_verifier.rs
@@ -338,11 +338,11 @@ where
         air.eval_folded_circuit(circuit, &sels, &alpha, &columns_targets, zero);
 
     let all_columns_targets = AllColumnTargets {
-        local_lookups,
-        global_lookups,
-        columns_targets: &columns_targets,
-        local_lookup_challenges: &challenge_targets.local_lookup_challenges,
-        global_lookup_challenges: &challenge_targets.global_lookup_challenges,
+        local_lookups: local_lookups.to_vec(),
+        global_lookups: global_lookups.to_vec(),
+        columns_targets,
+        local_lookup_challenges: challenge_targets.local_lookup_challenges,
+        global_lookup_challenges: challenge_targets.global_lookup_challenges,
     };
     // Fold the lookup constraints with the previously computed AIR constraints.
     let folded_constraints = lookup_gadget.eval_lookup_constraints(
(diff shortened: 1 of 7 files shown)


recursion/src/circuit_verifier.rs
```diff
@@ -82,13 +100,13 @@ pub fn verify_circuit<
     OpeningProof: Recursive<SC::Challenge>,
 >(
     config: &SC,
-    air: &A,
+    air: &dyn RecursiveAirWithLookupVerification<SC::Challenge, Comm>,
```
4l0n50, line 103: why do we need dynamic dispatch now?

recursion/src/lookup.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        folded_constraints = self.eval_local_folded_circuit(
+            circuit,
+            sels,
+            alpha,
+            &local_lookup_columns,
+            folded_constraints,
+        );
+
+        folded_constraints = self.eval_global_constraints(
```
4l0n50, line 141: don't you need to compute `folded_constrais = folded_constraints + alpha * self.eval_global_constraints(..)` in the circuit?

4l0n50 (commented):
Just took a quick look. It would be nice to see an example, but I guess it needs non-recursive lookups first.

recursion/src/lookup.rs
LindaGuiga: `eval_global_constraints` takes `folded_constraints` as an argument, so what it should do is use that start of the accumulator to keep folding

PR #120, opened 2025-10-02
mdbook: Expand IR section and detail tables

LindaGuiga:
I added some explanations in the IR part of the mdbook, and detailed the chips we currently have. Let me know whether there are too many details, and how I can improve these!

### DIFF
--- a/book/src/construction.md
+++ b/book/src/construction.md
@@ -35,13 +35,56 @@ An **Execution IR** (intermediate representation) is defined to describe the ste
 This IR is *not proved itself*; it only guides trace population.
 The actual soundness comes from the constraints inside the operation-specific STARK chips along with their lookups into the central witness table.
 
+In the IR, we differentiate between *primitive* and *nonprimitive operations*. 
+
+Primitive operations represent basic field arithmetic operations:
+
+- `Constant` -- for constants,
+- `Public` -- for public inputs,
+- `Add` -- for both addition and subtraction,
+- `Mul` -- for both multiplication and division. 
+
+They represent the core of the verification computation. Given only the primitive operations, one should be able to carry out most operations necessary in circuit verification. Primitive operations have the following properties:
+
+- They operate on elements of the witness table, through their `WitnessId` (index within the witness table).
+- The representation can be heavily optimized. For example, every time a constant is added to the IR, we wither create a new `WitnessId` or return an already existing one. We could also carry out common subexpression elimination.
+- They are executed in topological order during the circuit evaluation, and
+- they form a directed acyclic graph of dependencies.
+
+But relying only on primitive operations for the entire verification would lead to the introduction of many temporary values in the IR. In turn, this would lead to enlarged witness and primitive tables. This can be alleviated by the introduction of chips which carry out more complex operations: *nonprimitive operations*. These have more complex constraints, for example for sensitive cryptographic operations, and are verified by specialized AIRs. As such, they cannot be optimized similarly to primitive operations.
+
+These nonprimitive operations use not only witness table elements (including public inputs), but may also require the use of *private data*. For example, when verifying a Merkle path, hash outputs are not stored in the witness table. Nonprimitive operations make it easy to add *complex functionality*, and therefore contribute to the *modularity* of the system.
+
+In our implementation of the recursive verifier, we leverage the following nonprimitive operation:
+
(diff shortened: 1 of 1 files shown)


book/src/construction.md
```diff
@@ -35,13 +35,56 @@ An **Execution IR** (intermediate representation) is defined to describe the ste
 This IR is *not proved itself*; it only guides trace population.
 The actual soundness comes from the constraints inside the operation-specific STARK chips along with their lookups into the central witness table.
 
+In the IR, we differentiate between *primitive* and *nonprimitive operations*. 
+
+Primitive operations represent basic field arithmetic operations:
```
Nashtare, line 40: not really, for instance PIs are not arith ops per say

maybe rephrase in something like this?

```suggestion
Primitive operations represent the core of the verification computation, and will always be available when building a recursive verifier, regardless of the sub-protocols and parameterization. Non-primitive operations on the other-hand can be added, in a plug&play fashion, to optimize instructions that may be specific to particular usecases.

Some *primitive* operations are:  
```

(note that I reused your first sentence from above)

book/src/construction.md
```diff
@@ -35,13 +35,56 @@ An **Execution IR** (intermediate representation) is defined to describe the ste
...
+
+- `Constant` -- for constants,
+- `Public` -- for public inputs,
+- `Add` -- for both addition and subtraction,
+- `Mul` -- for both multiplication and division. 
+
+They represent the core of the verification computation. Given only the primitive operations, one should be able to carry out most operations necessary in circuit verification. Primitive operations have the following properties:
+
+- They operate on elements of the witness table, through their `WitnessId` (index within the witness table).
+- The representation can be heavily optimized. For example, every time a constant is added to the IR, we wither create a new `WitnessId` or return an already existing one. We could also carry out common subexpression elimination.
```
Nashtare, line 50:
```suggestion
- The representation can be heavily optimized. For example, every time a constant is added to the IR, we either create a new `WitnessId` or return an already existing one. We could also carry out common subexpression elimination.
```

book/src/construction.md
```diff
@@ -35,13 +35,56 @@ An **Execution IR** (intermediate representation) is defined to describe the ste
...
+- They are executed in topological order during the circuit evaluation, and
+- they form a directed acyclic graph of dependencies.
+
+But relying only on primitive operations for the entire verification would lead to the introduction of many temporary values in the IR. In turn, this would lead to enlarged witness and primitive tables. This can be alleviated by the introduction of chips which carry out more complex operations: *nonprimitive operations*. These have more complex constraints, for example for sensitive cryptographic operations, and are verified by specialized AIRs. As such, they cannot be optimized similarly to primitive operations.
```
Nashtare, line 54: actually you mention it here, so you can either tweak with my initial suggestion or ignore it :)

book/src/construction.md
```diff
@@ -55,6 +98,87 @@ A chip contains:
...
 
+**Primitive operations**: 
+
+Note that in the following tables, all index columns are preprocessed.
+
+- `WitnessAir`:
+
+  - It is the central AIR in our construction. It represents the global witness bus: it stores all witness indices and associated values used in the verifier circuit.
+
+  - ***columns***: the AIR has `D + 1` columns. They correspond to: `1` column for the index of the element, and `D` columns to store the value. Note that in this table, indices are incremented at each row.
```
Nashtare, line 109: I'm a bit skeptical about this approach of listing in the doc explicit mentions of the columns and constraints. What the type 1 doc told us is that this gets out of date **extremely** fast.

I'd rather have these as rustdoc within the air files themselves at the top, especially for auditors, but not as part of the book, or only a high level

sai-deng: What about using a toy example (e.g. the one from the design doc) to explain an end-to-end system architecture, and providing actual file links for real implementation details?

book/src/construction.md
LindaGuiga: I updated it to be the first sentence of this suggestion (since as you said below, the rest is mentioned later on)

book/src/construction.md
Nashtare: I think it ends up looking now like what I've done yesterday in #124, no? I guess we'll probably merge both PRs

Nashtare: Yeah sounds good to me

PR #121, opened 2025-10-02
Add run examples in ci

sai-deng:


Nashtare (commented):
maybe we could just add `--all-targets` instead to the job above, that way we only have 1 CI trigger, and it encompasses everything (test, benches, examples, ...)

sai-deng:
> maybe we could just add `--all-targets` instead to the job above, that way we only have 1 CI trigger, and it encompasses everything (test, benches, examples, ...)

Done. I also want to run the examples in CI.

Nashtare (approved):

PR #122, opened 2025-10-05
feat: add basic tracing and parallel support

Nashtare:
Added tracing logging for the `Fibonacci` example, and also added `parallel` feature to `circuit-prover` crate to leverage `rayon` from p3.


circuit-prover/src/config.rs
```diff
@@ -87,7 +87,7 @@ where
     let challenge_mmcs = ChallengeMmcs::<F, P, CD>::new(val_mmcs.clone());
 
     let dft = Dft::<F>::default();
-    let fri_params = create_test_fri_params::<ChallengeMmcs<F, P, CD>>(challenge_mmcs, 0);
+    let fri_params = create_benchmark_fri_params::<ChallengeMmcs<F, P, CD>>(challenge_mmcs);
```
Nashtare, line 90: unrelated, but the test config only has 2 queries, I think it's best to use a regular config from the start. Doesn't affect initial proving anyway

PR #126, opened 2025-10-06
doc: Add header rustdoc to primitive chips

LindaGuiga:


circuit-prover/src/air/mul_air.rs
```diff
@@ -1,3 +1,39 @@
+//! [`MulAir`] deals with multiplication and division. In the case of division, `a / b = c` is written in the table as `b * c = a`.
+//! The chip handles both base field and extension field operations, as it is parametrized by the extensions degree `D`.
```
Nashtare, line 2:
```suggestion
//! The chip handles both base field and extension field operations, as it is parametrized by the extension degree `D`.
```

circuit-prover/src/air/witness_air.rs
```diff
@@ -1,3 +1,26 @@
...
+//! Each row in the table increments the index by 1, starting from 0 (and so the index column is the range from 0 to `height - 1`).
+//!
+//!  # Constraints
+//!  - in the first row: ensure that `index = 0`.
+//!  - for transitions: `index_next - index_current - 1`.
+//!
+//! # Global Interactions
+//!
+//! Since this AIR serves as a witness bus, where the other chips read values from, it has interactions with all the other
+//! (primitive operation) AIRs.
```
Nashtare, line 21:
```suggestion
//! Since this AIR serves as a witness bus, where the other chips read values from, it has interactions with all the other
//! AIRs.
```

I believe non-primitive operations may also communicate with the Witness bus?

Nashtare (approved):
Overall looking good! Only minor comments

circuit-prover/src/air/witness_air.rs
LindaGuiga: I was thinking that nonprimitive operations *may* not have interactions with it but primitive operations always do. But I think you're right that it would be easier to just say it has interactions with everything.

PR #127, opened 2025-10-06
doc: Explain how we deduce an AIR's circuit constraints

LindaGuiga:


book/src/circuit_building.md
```diff
@@ -54,6 +54,54 @@ be pruned away from the graph, and the operation removed.
...
 
+## Building recursive AIR constraints
+
+In order to recursively verify an AIR, its constraints need to be added to the circuit and folded together. In our Fibonacci example, we had to add the constraints manually, one by one. But in reality, it is possible to automatically deduce the circuit constraints from its original set of constraints by leveraging Plonky3's symbolic constraint representation.
+
+In plonky3, the method `get_symbolic_constraints` for Airs returns a symbolic representation of constraints. Since our primitive chips (see section "Execution IR") encompass the various entries in the symbolic representation, we can map the operations to their circuit equivalent automatically. The `symbolic_to_circuit` function does exactly that. 
+
+We can explain in more details how the automated deduction of constraints and folding work. Assume we have a circuit builder `builder`, an `ExprId` for folding challenge `alpha`, and `ExprId`s for the local row (`local`) and the next row (`next`) of a `FibonacciAir`. `FibonacciAir`'s constraints for public inputs `a`, `b` and `output` are the following:
```
hratoanina, line 63: I don't think we really need to explain the internals. For me, just saying that we can automatically convert P3 symbolic constraints into a circuit, and adding a concrete example in code is enough.

Nashtare: Agree with Hamy, or at least make the explanation generic enough as the current seems fairly verbose for little reason. Making keeping the big lines as to how symbolic constraints are being internally mapped to equivalent expressions that are then propagated to their associated primitive chips is enough? Or something like this

PR #131, opened 2025-10-09
Add local poseidon2-air crate with extra index columns

sai-deng:
The local copy will be extended to verify Poseidon2 operations used in challenger/sponge circuits and MMCS verification AIR with CTLs.

### DIFF
--- a/poseidon2-circuit-air/Cargo.toml
+++ b/poseidon2-circuit-air/Cargo.toml
@@ -11,9 +11,9 @@ categories.workspace = true
 
 [dependencies]
 p3-air.workspace = true
+p3-baby-bear.workspace = true
 p3-circuit.workspace = true
 p3-field.workspace = true
-p3-baby-bear.workspace = true
 p3-koala-bear.workspace = true
 p3-matrix.workspace = true
 p3-maybe-rayon.workspace = true


circuit/src/tables/poseidon2.rs
```diff
@@ -0,0 +1,10 @@
+/// Poseidon2 operation table
```
sai-deng: Should we have the state and flags in this struct?

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,266 @@
+use core::array;
```
sai-deng: I suggest rename it to p3-poseidon2-recursion-air as poseidon2-circuit-air does not emphasize the extra columns you add for the circuit

poseidon2-circuit-air/src/columns.rs
```diff
@@ -0,0 +1,152 @@
...
+use p3_poseidon2_air::Poseidon2Cols;
+
+/// Columns for a Poseidon2 AIR which computes one permutation per row.
+///
+/// They extend the P3 columns with some circuit-specific columns.
+///
+/// `reset` (transparent): indicates whether the state is being reset this row.
+/// `absorb_flags` (transparent): for each rate element, indicates if it is being absorbed this row.
+/// At most one flag is set to 1 per row: if `absorb_flags[i]` is 1, then all elements up to the `i`-th
+/// are absorbed; the rest are propagated from the previous row.
```
sai-deng: I think we do not need so many flags? The use cases are either first half or all?

sai-deng (commented):
Could we include a minimal trace-generation helper plus unit tests to harden this PR?

poseidon2-circuit-air/src/sub_builder.rs
```diff
@@ -0,0 +1,108 @@
+// Code from SP1 with minor modifications:
+// https://github.com/succinctlabs/sp1/blob/main/crates/stark/src/air/sub_builder.rs
+
+use alloc::vec::Vec;
+use core::ops::{Deref, Range};
+
+use p3_air::{Air, AirBuilder};
+use p3_matrix::Matrix;
+
+/// A submatrix of a matrix.  The matrix will contain a subset of the columns of `self.inner`.
```
Nashtare:
```suggestion
/// A submatrix of a matrix. The matrix will contain a subset of the columns of `self.inner`.
```

Nashtare: also nit but doesn't submatrix imply it's from a "larger" matrix already?

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,676 @@
...
+        SBOX_DEGREE,
+        SBOX_REGISTERS,
+        HALF_FULL_ROUNDS,
+        PARTIAL_ROUNDS,
+    >
+{
+    pub const fn new(
+        constants: RoundConstants<F, WIDTH, HALF_FULL_ROUNDS, PARTIAL_ROUNDS>,
+    ) -> Self {
+        assert!(CAPACITY_EXT + RATE_EXT == WIDTH_EXT);
```
4l0n50: You can also do
```rust
const _: () = {
            assert!(CAPACITY_EXT + RATE_EXT == WIDTH_EXT, "capacity+rate mismatch");
            assert!(WIDTH_EXT * D == WIDTH, "width extension mismatch");
        };
```

poseidon2-circuit-air/src/sub_builder.rs
```diff
@@ -0,0 +1,108 @@
...
+    #[inline]
+    fn height(&self) -> usize {
+        self.inner.height()
+    }
+}
+
+/// A builder used to eval a sub-air.  This will handle enforcing constraints for a subset of a
+/// trace matrix.  E.g. if a particular air needs to be enforced for a subset of the columns of
+/// the trace, then the [`SubAirBuilder`] can be used.
+pub struct SubAirBuilder<'a, AB: AirBuilder, SubAir: Air<AB>, T> {
```
Nashtare: FYI following yesterday's discussion, I tried tweaking to `BaseAir` and didn't have any issues with it, not sure what you faced (3 lines change in total)

Nashtare:
Fyi @hratoanina I've pulled in latest main for CI to go ✅

poseidon2-circuit-air/src/air.rs
Nashtare: nit, but just mentioning `const` is enough:


```rust
const {
    assert!(CAPACITY_EXT + RATE_EXT == WIDTH_EXT, "capacity+rate mismatch");
    assert!(WIDTH_EXT * D == WIDTH, "width extension mismatch");
}
```

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,680 @@
...
+        type U64Hash = PaddingFreeSponge<KeccakF, 25, 17, 4>;
+        let u64_hash = U64Hash::new(KeccakF {});
+
+        type FieldHash = SerializingHasher<U64Hash>;
+        let field_hash = FieldHash::new(u64_hash);
+
+        type MyCompress = CompressionFunctionFromHasher<U64Hash, 2, 4>;
+        let compress = MyCompress::new(u64_hash);
+
+        // WARNING: DO NOT USE SmallRng in proper applications! Use a real PRNG instead!
```
Nashtare: can probably be removed 🙃

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,680 @@
...
+    fn width(&self) -> usize {
+        num_cols::<
+            WIDTH_EXT,
+            RATE_EXT,
+            Poseidon2Cols<u8, WIDTH, SBOX_DEGREE, SBOX_REGISTERS, HALF_FULL_ROUNDS, PARTIAL_ROUNDS>,
+        >()
+    }
+}
+
+pub(crate) fn eval<
```
Nashtare: probably shouldn't even be `pub(crate)`

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,680 @@
...
+            D,
+            WIDTH,
+            WIDTH_EXT,
+            RATE_EXT,
+            CAPACITY_EXT,
+            SBOX_DEGREE,
+            SBOX_REGISTERS,
+            HALF_FULL_ROUNDS,
+            PARTIAL_ROUNDS,
+        > = Poseidon2CircuitAir::new(constants.clone());
```
Nashtare: similarly to the above suggestion, would be nice if you were defining some types for commonly used configurations to make it lighter for users, maybe in some `public_types.rs` module or similar

poseidon2-circuit-air/src/columns.rs
```diff
@@ -0,0 +1,79 @@
...
+/// `sponge_reset`: auxiliary column to keep constraint degrees below three.
+/// `absorb_flags` (transparent): for each rate element, indicates if it is being absorbed this row.
+/// At most one flag is set to 1 per row: if `absorb_flags[i]` is 1, then all elements up to the `i`-th
+/// are absorbed; the rest are propagated from the previous row.
+/// `input_indices` (transparent): for each input element, indicates the index in the witness table for the
+/// memory lookup. It's either received (for an absorb or a compression) or sent (for a squeeze).
+/// `output_indices` (transparent): for each output element, indicates the index in the witness table for the
+/// memory lookup. Only used by compressions to send the output.
+#[repr(C)]
+pub struct Poseidon2CircuitCols<
```
Nashtare: Plonky3 uses a `vectorized` version of Poseidon2 air for optimal performances (prover side) with a default value of 8. While it's unclear with recursion what the best value is, it may be worth investigating the tradeoff and possibly upgrade the AIR here to account for this. Perhaps we can add a TODO for that?

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,680 @@
...
+        extra_capacity_bits: usize,
+        perm: P,
+    ) -> RowMajorMatrix<F> {
+        let n = sponge_ops.len();
+        assert!(
+            n.is_power_of_two(),
+            "Callers expected to pad inputs to a power of two"
+        );
+
+        let num_circuit_cols = 3 + 2 * RATE_EXT + WIDTH_EXT;
```
Nashtare: nit: can be made const
```suggestion
        const NUM_COLS: usize = 3 + 2 * RATE_EXT + WIDTH_EXT;
```

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,680 @@
...
+                .expect("Missing row {i}?")
+                .into_iter()
+                .collect::<Vec<_>>();
+            row[..p2_ncols].copy_from_slice(&left_part);
+            row[p2_ncols..].copy_from_slice(&right_part);
+        }
+
+        unsafe {
+            vec.set_len(n * ncols);
+        }
```
Nashtare:
```suggestion
```

this is already guaranteed

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,680 @@
...
+        assert!(
+            n.is_power_of_two(),
+            "Callers expected to pad inputs to a power of two"
+        );
+
+        let num_circuit_cols = 3 + 2 * RATE_EXT + WIDTH_EXT;
+        let mut circuit_trace = vec![F::ZERO; n * num_circuit_cols];
+        let mut circuit_trace = RowMajorMatrixViewMut::new(&mut circuit_trace, num_circuit_cols);
+
+        // TODO: Put trace generation circuit-side.
```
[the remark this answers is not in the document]
hratoanina: I meant that the work of computing the capacity for each row will be done circuit-side when executing it, and the events will contain the capacity values directly.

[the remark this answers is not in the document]
Nashtare: Why can't this be done in this PR right away?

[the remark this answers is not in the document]
Nashtare: I'd tend to think this is actually kind of against what we were advocating when discussing iterations with Sai on his own PR. IMO we should be advocating for features instead of mere code. Merging code that adds a placeholder and no functional gain just stores up technical debt and we pay the price on the long term.

Instead I think we should be asking

> What functionality does this PR bring?

Here, for now this is adding no functionality at all. The same issue happened with the previous PR you authored #59 which ends up being unused / possibly deprecated, which could have been different or at least anticipated if it had brought an actual complete feature by being leveraged in a part of the system.

The same issue happens here, and it would be worthwhile adding support in the circuit so that:

- we don't spend time implementing stuff we _know_ we won't need (cf. Sai's point)
- we actually can test the newly added functionality, with the behavior _it is intended to have_ in the first place

If the circuit‑side path is the chosen solution, I believe it should be done now so we don’t have to write and remove trace‑generation code twice, and keep on delaying delivery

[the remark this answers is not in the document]
hratoanina: I agree in general, the issue is that trace generation on the circuit side is still unstable, and very much WIP, using either Alonso's hint approach or the primitive/non-primitive op unification. In hindsight it should have been better to finish the work on the circuit side before working on tables, but in the current state finishing that work on this PR sounds majorly out of scope.

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,623 @@
...
+        Poseidon2Cols<
+            AB::Var,
+            WIDTH,
+            SBOX_DEGREE,
+            SBOX_REGISTERS,
+            HALF_FULL_ROUNDS,
+            PARTIAL_ROUNDS,
+        >,
+    >,
+) {
```
sai-deng: Some missing constraints (to enforce bool values and more):
```
    // Booleanity constraints for control columns and absorb flags (local and next).
    // x in {0,1} => x * (1 - x) = 0.
    builder.assert_bool(local.is_sponge.clone());
    builder.assert_bool(local.reset.clone());
    builder.assert_bool(local.sponge_reset.clone());
    builder.assert_bool(next.is_sponge.clone());
    builder.assert_bool(next.reset.clone());
    builder.assert_bool(next.sponge_reset.clone());

    // reset => is_sponge
    builder.assert_zero(local.reset.clone() * (AB::Expr::ONE - local.is_sponge.clone()));

    // Absorb flags booleanity and "at most one set" per row.
    let mut local_flags_sum = AB::Expr::ZERO;
    let mut next_flags_sum = AB::Expr::ZERO;
    for i in 0..RATE_EXT {
        builder.assert_bool(local.absorb_flags[i].clone());
        builder.assert_bool(next.absorb_flags[i].clone());
        local_flags_sum = local_flags_sum + local.absorb_flags[i].clone();
        next_flags_sum = next_flags_sum + next.absorb_flags[i].clone();
    }
    // sum in {0,1}
    builder.assert_bool(local_flags_sum);
    builder.assert_bool(next_flags_sum);

    // For compression rows, enforce all absorb flags are 1 (we omit the is_sponge check elsewhere
    // assuming this property). This keeps semantics consistent with comments.
    let ones_minus_flags_local = array::from_fn::<_, RATE_EXT, _>(|i| {
        AB::Expr::ONE - local.absorb_flags[i].clone()
    });
    builder
        .when(AB::Expr::ONE - local.is_sponge.clone())
        .assert_zeros(ones_minus_flags_local);

```

poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,623 @@
...
+    );
+    // During a reset, the rate elements not being absorbed are zeroed.
+    for (i, col) in current_no_absorb.iter().enumerate() {
+        let arr = array::from_fn::<_, D, _>(|j| local.poseidon2.inputs[i * D + j].clone().into());
+        builder
+            .when(local.sponge_reset.clone() * col.clone())
+            .assert_zeros(arr);
+    }
+
+    let _is_squeeze = AB::Expr::ONE - current_absorb[0].clone();
```
sai-deng: I think you may need something like this
```
    let mut sum_flags = AB::Expr::ZERO;
    for i in 0..RATE_EXT {
        sum_flags = sum_flags + local.absorb_flags[i].clone();
    }
    let _is_squeeze = AB::Expr::ONE - sum_flags;
```

poseidon2-circuit-air/src/air.rs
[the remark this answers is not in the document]
Nashtare: I would still disagree to be honest. For now, this PR adds code that isn't integrated in any other part of the system, which should be a no-go to begin with. This adds bloat and tech debt.

PR #148, opened 2025-10-13
Validate proof shapes and PoW

### LINKED ISSUE #30: Validate inner STARK shape/params
Perform basic checks, i.e. check widths, degrees, blowup, etc... before verification.

LindaGuiga:
Should close #30

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -4,6 +4,7 @@
 * [Introduction](introduction.md)
 * [Construction](construction.md)
 * [Circuit building](circuit_building.md)
+* [Trace generation](trace_generation.md)
 * [Handling arbitrary programs](extensions.md)
 * [Debugging](debugging.md)
 * [Benchmarks](benchmark.md)
--- a/book/src/trace_generation.md
+++ b/book/src/trace_generation.md
@@ -1,0 +1,119 @@
+# Trace Generation Pipeline
+
+After building a circuit, the next step is execution: running the program with concrete inputs and generating the execution traces needed for proving. This section describes the complete flow from a static `Circuit` specification to the final `Traces` structure.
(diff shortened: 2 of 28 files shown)


recursion/src/recursive_challenger.rs
```diff
@@ -34,6 +35,35 @@ pub trait RecursiveChallenger<F: Field> {
...
     }
 
+    fn sample_bits(
+        &mut self,
+        circuit: &mut CircuitBuilder<F>,
+        total_num_bits: usize,
+        num_bits: usize,
+    ) -> Vec<Target> {
+        let x = self.sample(circuit);
+        // Adds one public input per bit (there are as many bits as for a field element).
```
Nashtare: perhaps to signal the bits are expected to be public inputs
```suggestion
    fn sample_public_bits(
        &mut self,
        circuit: &mut CircuitBuilder<F>,
        total_num_bits: usize,
        num_bits: usize,
    ) -> Vec<Target> {
        let x = self.sample(circuit);
        // Adds one public input per bit (there are as many bits as for a field element).
```

recursion/src/recursive_generation.rs
```diff
@@ -244,7 +247,25 @@ where
         }
         // Observe PoW and sample bits.
         let pow_bits = params[0];
-        if !challenger.check_witness(pow_bits, opening_proof.pow_witness) {
+        // Check PoW witness.
```
sai-deng: use challenger.check_witness() here?

sai-deng: I am wondering if we need PoW payload into challenges.
In `recursive_pcs.rs::verify_circuit` you slice challenges as if there were no PoW payload between betas and query indices.

Issue #147, opened 2025-10-12
Refactor `CircuitBuilder` into specialized sub-components


Nashtare:
### Issue

I have currently two main issues with how the `CircuitBuilder` is implemented, which I think are worth discussing:

- the module is huge (~1750 LoC at the moment) 
- it currently holds a fairly complex logic, all the way from statement definition to IR lowering and final circuit construction, which seems better suited to distinct, specialized sub-components.

### Suggestion

Split the `CircuitBuilder` logic into separate specialized builder entities, each responsible for a portion of the overall logic.
This:

- makes each dedicated module more developer friendly (no huge unique file)
- allows for separation of concern, with each sub-builder executing only its own specific piece of logic. This would give us:
  - no cross-dependencies
  - better and simpler management of each component
  - better composability

Ideally, I think we may want something like this (names arbitrary)

```rust
pub struct CircuitBuilder<F> {
    expr_builder: ExpressionBuilder<F>,
    witness_allocator: WitnessAllocator,
    public_tracker: PublicInputTracker, 
    non_primitive_ops: Vec<(NonPrimitiveOpId, NonPrimitiveOpType, Vec<ExprId>)>,
    config: BuilderConfig,
}

pub struct CircuitCompiler<F> {
    expr_lowerer: ExpressionLowerer<F>,
    non_primitive_lowerer: NonPrimitiveLowerer<F>,
    optimizer: PrimitiveOptimizer<F>,
    config: CircuitConfig,
}
```

where for the `CircuitBuilder` fields:

- `expression_builder` handles the whole expression graph construction
- `witness_allocator` doesn't really change
-  `public_tracker` declares and keeps track of the public inputs
- `non_primitive_builder` 
- `config` handles all the compile-time/runtime info of our circuit, for now corresponding to the set of enabled non-primitive operations. It could also contain optimization-related arguments, which depending on the target goal for the circuit being built could for instance do additional optimization passes w.r.t non-primitive ops or hint at an optimized `TablePacking` config?

and for `CircuitCompiler` (the one responsible to do the whole lowering steps and optimization passes:

- a specialized `_lowerer` for each component with dedicated logic
- an `optimizer` to perfom all the optimization passes
- `config`

PR #149, opened 2025-10-13
circuit builder: refactor using specialized sub-components

tcoratger:
Should close https://github.com/Plonky3/Plonky3-recursion/issues/147

@Nashtare let me know what you think about it.

Some functions are smaller or broken into several pieces so that the unit tests are not exactly the same as the previous ones but I've tried to cover everything as best as possible.

### DIFF
--- a/circuit/src/builder/compiler/non_primitive_lowerer.rs
+++ b/circuit/src/builder/compiler/non_primitive_lowerer.rs
@@ -2,8 +2,7 @@ use alloc::vec::Vec;
 
 use hashbrown::HashMap;
 
-use crate::builder::CircuitBuilderError;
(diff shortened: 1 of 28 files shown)


circuit/src/builder/compiler/non_primitive_lowerer.rs
```diff
@@ -0,0 +1,462 @@
+use alloc::vec::Vec;
+
+use hashbrown::HashMap;
+
+use crate::builder::CircuitBuilderError;
+use crate::builder::config::BuilderConfig;
```
sai-deng: nit:
use crate::builder::{BuilderConfig, CircuitBuilderError};

PR #150, opened 2025-10-14
refactor: non-primitive handling of witness data

Nashtare:
Following the discussion we had highlighting the current issues of having to provide all necessary inputs for non-primitive ops to the `CircuitRunner` before the actual run, here's my attempt at simplifying the barrier between `Primitive` vs `NonPrimitive` ops, as well as allowing a custom execution context for non-primitive ops to alleviate the issue above.

For now the way we pass private witness data is unchanged, but this should also ease a transition towards automatic handling at runtime, so that users don't need to pre-generate specific data for each instance of their non-primitive ops.

Main changes:
- merged all operations together, renaming `Prim` to `Op` to reflect it
- expression lowering now handles both kinds, although it outputs separate vectors for convenience (for future optimization passes and for faster iteration when filling up auxiliary traces)
- mapped to the (new) `Expr::NonPrimitiveOp` is an `Op::NonPrimitiveOpWithExecutor`, which contains
  - inputs (as `Vec<ExprId>`)
  - outputs (as `Vec<ExprId>`)
  - an associated `ExprId` (unlike the initial design not assigning any to non-primitive operations)
  - an `executor`, which defines custom logic specific to the operation to interact with the runner (so that it can get/set witnesses, execute its own internal logic and interact with private data (if needed))
  
There is still _more_ work to allow for proper dynamic dispatch at runtime with minimum disruption but this isn't the main goal of this PR so I'd rather hold off any other changes.

### DIFF
--- a/circuit/src/builder/compiler/non_primitive_lowerer.rs
+++ b/circuit/src/builder/compiler/non_primitive_lowerer.rs
@@ -63,7 +63,7 @@ impl<'a> NonPrimitiveLowerer<'a> {
                     };
 
                     let ext = config.ext_field_digest_elems;
-                    let expected = ext + 1 + ext; // leaf (ext) + index (1) + root (ext)
+                    let expected = config.input_size();
                     if witness_exprs.len() != expected {
                         return Err(CircuitBuilderError::NonPrimitiveOpArity {
                             op: "MmcsVerify",
@@ -102,7 +102,7 @@ impl<'a> NonPrimitiveLowerer<'a> {
                         .collect::<Result<_, _>>()?;
 
                     // Build Op with executor: all in inputs; no outputs
-                    let mut inputs = leaf_widx.clone();
+                    let mut inputs = leaf_widx;
                     inputs.push(index_widx);
                     inputs.extend(root_widx);
                     lowered_ops.push(Op::NonPrimitiveOpWithExecutor {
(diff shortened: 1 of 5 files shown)


circuit/src/builder/expression_builder.rs
```diff
@@ -98,6 +98,29 @@ where
         expr_id
     }
 
+    /// Adds a witness hint to the graph.
+    /// It will allocate a `WitnessId` during lowering, with no primitive op.
+    #[allow(unused_variables)]
```
4l0n50:
```suggestion
```

circuit/src/builder/expression_builder.rs
```diff
@@ -98,6 +98,29 @@ where
...
+    pub fn add_witness_hint(&mut self, label: &'static str) -> ExprId {
+        let expr_id = self.graph.add_expr(Expr::Witness);
+
+        #[cfg(debug_assertions)]
+        self.allocation_log.push(AllocationEntry {
+            expr_id,
+            alloc_type: AllocationType::WitnessHint,
+            label,
```
4l0n50:
```suggestion
            _label,
```

Nashtare: Underscore implies it's never used, this is not necessarily the case

PR #153, opened 2025-10-16
Adapt the Mmcs gate interface

4l0n50:
In order to use the recursive version of the [verify_batch function](https://github.com/Plonky3/Plonky3/blob/6aa45e3eae911f2f6f591682a7ccae480105c57d/commit/src/mmcs.rs#L166), we need to open the commitment to multiple leaves (I used “leaves” everywhere, but writing this now I realize it may not be the right term).

On the other hand, in the recursive verification of FRI proofs, the bit decomposition of the index is already available. Therefore, the MMCS gate might avoid this conversion and work directly with the index bits.

This PR updates the MMCS gate to support multiple openings. To enable this, non-primitive operations now expose `witness_exprs` as `Vec<Vec<ExprId>>`. In the MMCS case,  the last two vectors hold directions and root; the preceding |directions| vectors contain the openings for each matrix (previously referred to as “leaves,” possibly incorrectly), or an empty vector if there's no matrix at the respective level.

### DIFF
--- a/circuit-prover/examples/mmcs_verify.rs
+++ b/circuit-prover/examples/mmcs_verify.rs
@@ -47,7 +47,7 @@ fn main() -> Result<(), ProverError> {
     // we're proving the opening of an Mmcs to matrices of height 2^depth, 2^(depth -1), ...
     let leaves: Vec<Vec<ExprId>> = (0..depth)
         .map(|i| {
-            (0..if (i as usize).is_multiple_of(2) && i != depth - 1 {
+            (0..if i % 2 == 0 && i != depth - 1 {
                 mmcs_config.ext_field_digest_elems
             } else {
                 0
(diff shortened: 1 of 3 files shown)


circuit/src/tables/mmcs.rs
```diff
@@ -135,29 +151,54 @@ impl<F: Field + Clone + Default> MmcsPrivateData<F> {
...
+        // In the first iteration the leaf is assigned to the state, consequently
+        // there's no
+        // We need an empty leaf to replace the one we already assigned to the state;
+        let empty_leaf = vec![];
+        for (&dir, sibling, leaf) in izip!(
+            directions.iter(),
+            siblings.iter(),
+            iter::once(&empty_leaf).chain(leaves.iter().skip(1))
+        ) {
+            path_states.push((
```
sai-deng: I think the order is not correct? Besides, the mock compression is to simple to catch bugs, could you please change mock compression to something like:
        let perm = default_babybear_poseidon2_16();
        let compress: TruncatedPermutation<Poseidon2BabyBear<16>, 2, 1, 16> = TruncatedPermutation::new(perm);

sai-deng (commented):
nvm

sai-deng:
Previously: we’d call MmcsVerify multiple times, bottom→up, ending a segment whenever we hit a new leaf layer. I am wondering how current method is better than the old method?

PR #154, opened 2025-10-17
book: add trace generation section

tcoratger:


book/src/trace_generation.md
```diff
@@ -0,0 +1,120 @@
...
+Trace extraction happens internally within `runner.run()` after execution completes. This phase delegates to specialized trace builders that transform the populated witness table into operation-specific trace tables.
+
+Each operation family has a dedicated builder that extracts its operations from the IR and produces trace columns:
+
+- **WitnessTraceBuilder** — Generates the central [witness table](./construction.md#witness-table) with sequential `(index, value)` pairs
+- **ConstTraceBuilder** — Extracts constants (both columns preprocessed)
+- **PublicTraceBuilder** — Extracts public inputs (index preprocessed, values at runtime)
+- **AddTraceBuilder** — Extracts additions with six columns: `(lhs_index, lhs_value, rhs_index, rhs_value, result_index, result_value)`
+- **MulTraceBuilder** — Extracts multiplications with the same six-column structure
+- **MmcsTraceBuilder** — Validates and extracts MMCS path verification traces
```
LindaGuiga: Maybe we should just list the primitive ops, as it might be difficult to maintain if we want to add NonPrimitiveOps? Also, since NonPrimitiveOps are supposed to be modular, people will have to create their own TraceBuilder... So maybe we could list the ones for PrimitiveOps and just mention MmcsTraceBuilder as an example of how to create one's TraceBuilder?

LindaGuiga (approved):
Thank you!

PR #155, opened 2025-10-22
Add compress table

hratoanina:


poseidon2-compress-air/src/air.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use crate::{Poseidon2CompressCols, num_cols};
+
+/// Extends the Poseidon2 AIR with recursion circuit-specific columns and constraints.
+/// Assumes the field size is at least 16 bits.
+///
+/// It compresses `N` chunks of size 'CHUNK' of degree-`D` extension field elements into one chunk.
+#[derive(Debug)]
+pub struct Poseidon2CompressAir<
+    F: PrimeCharacteristicRing,
```
4l0n50: Is it possible to encapsulate all these consts in a single config trait?

poseidon2-compress-air/src/air.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use p3_air::{Air, AirBuilder, BaseAir};
+use p3_field::PrimeCharacteristicRing;
+use p3_matrix::Matrix;
+use p3_poseidon2::GenericPoseidon2LinearLayers;
+use p3_poseidon2_air::{Poseidon2Air, RoundConstants};
+
+use crate::{Poseidon2CompressCols, num_cols};
+
+/// Extends the Poseidon2 AIR with recursion circuit-specific columns and constraints.
```
4l0n50:
```suggestion
/// Extends the Poseidon2 AIR with recursion circuit-specific columns and lookups.
```
Because I can't see any change in the constraints.

hratoanina:
Yes, there's no new constraint in that table before lookups, but I think it's still worth it to have it now so that we can at least add the table and connect it to a new `compress` op.

sai-deng:
Per our offline discussion, let’s close this PR for now since we don’t need two Poseidon tables in the recursion proof?

Issue #46, opened 2025-08-30
Add `zk` as an optional recursion parameter


Nashtare:
Initial recursion implementation most likely will scrap off zk for simplicity. It may be nice to support it later-on.

Nashtare:
Implemented in #328

PR #159, opened 2025-10-29
Add Recursive Batch-STARK Verifier Circuits

sai-deng:
Also removed MMCS from the batch proof.

### DIFF
--- a/recursion/src/verifier/batch_stark.rs
+++ b/recursion/src/verifier/batch_stark.rs
@@ -70,7 +70,7 @@ where
     }
 }
 
-/// Build and attach a recursive verifier circuit for a circuit-prover BatchStarkProof.
+/// Build and attach a recursive verifier circuit for a circuit-prover [`BatchStarkProof`].
 ///
 /// This reconstructs the circuit table AIRs from the proof metadata (rows + packing) so callers
 /// don't need to pass `circuit_airs` explicitly. Returns the allocated input builder to pack


Nashtare (commented):
Maybe worth renaming as it also contains the removal of MMCS?

recursion/src/verifier/batch_stark.rs
```diff
@@ -0,0 +1,561 @@
...
+        )),
+        CircuitTablesAir::Mul(MulAir::<SC::Challenge, TRACE_D>::new(
+            rows[Table::Mul],
+            mul_lanes,
+        )),
+    ];
+
+    // TODO: public values are empty for all circuit tables for now.
+    let air_public_counts = vec![0usize; proof.proof.opened_values.instances.len()];
+    let verifier_inputs = crate::public_inputs::BatchStarkVerifierInputsBuilder::<
```
Nashtare:
```suggestion
    let verifier_inputs = BatchStarkVerifierInputsBuilder::<
```

recursion/src/verifier/batch_stark.rs
```diff
@@ -0,0 +1,561 @@
...
+            SC,
+            InputProof,
+            OpeningProof,
+            Comm,
+            <SC::Pcs as Pcs<SC::Challenge, SC::Challenger>>::Domain,
+        >,
+    SC::Challenge: PrimeCharacteristicRing,
+    <<SC as StarkGenericConfig>::Pcs as Pcs<SC::Challenge, SC::Challenger>>::Domain: Clone,
+{
+    debug_assert_eq!(config.is_zk(), 0, "batch recursion assumes non-ZK");
```
Nashtare: Is this set in stone? I thought we wanted to add ZK support later on? If so, then I'd rather have a TODO associated with the existing issue (#46) for clarity

recursion/src/verifier/batch_stark.rs
```diff
@@ -0,0 +1,561 @@
...
+            "ZK mode is not supported for batch-STARK recursion".to_string(),
+        ));
+    }
+
+    // Enable hash operations for CircuitChallenger
+    circuit.enable_op(
+        NonPrimitiveOpType::HashAbsorb { reset: true },
+        NonPrimitiveOpConfig::None,
+    );
+    circuit.enable_op(NonPrimitiveOpType::HashSqueeze, NonPrimitiveOpConfig::None);
```
Nashtare: I find this kinda dangerous / ugly to be plugged in manually / hardcoded within the function itself.

Philosophically, non-primitive ops should all be opt-in with some eventual predetermined configs based on common setups including relevant ones, but not half-backed in like so.

Nashtare (changes requested):
To reply to your last comment

> It did not completely remove MMCS. Just removed it from BatchProof to make the circuit simpler.

I find it quite confusing tbh. Either we remove MMCS altogether in this PR or we keep it in both regular and batch proof until future removal. Also note that it makes your approach _look_ simpler, but that's because you're abstracting away the handling of non-primitive ops at runtime.

sai-deng:
> To reply to your last comment
> 
> > It did not completely remove MMCS. Just removed it from BatchProof to make the circuit simpler.
> 
> I find it quite confusing tbh. Either we remove MMCS altogether in this PR or we keep it in both regular and batch proof until future removal. Also note that it makes your approach _look_ simpler, but that's because you're abstracting away the handling of non-primitive ops at runtime.

My goal is to keep this PR as simple as possible. Why are we adding it to the recursion circuits if this table will be removed soon?

Nashtare:
You could always do it on a separate PR targeting this branch, so that both can be reviewed / merged together before landing into `main`. That keeps PRs light, without having the inconsistency that the current PR brings to `main`.

We must strive to keep the codebase as clean and understandable as possible for external developers too. I don't think the argument "I removed MMCS only here because it makes things simpler" is valid. In that case, we can first merge a PR removing it altogether, and then we review this PR, which will remain clean and light?

sai-deng:
> You could always do it on a separate PR targeting this branch, so that both can be reviewed / merged together before landing into `main`. That keeps PRs light, without having the inconsistency that the current PR brings to `main`.
> 
> We must strive to keep the codebase as clean and understandable as possible for external developers too. I don't think the argument "I removed MMCS only here because it makes things simpler" is valid. In that case, we can first merge a PR removing it altogether, and then we review this PR, which will remain clean and light?

I disagree. MMCS is currently the only non-primitive operation in the system, and I think it would be better to refactor it into circuits rather than fully removing it right now, to avoid introducing unnecessary simplifications. Also, we can’t really keep the codebase clean at this moment, since the MMCS refactor is blocked by the Poseidon table work.

Nashtare:
Just to clarify, by fully removing it I meant removing the dedicated table, not the logic. Removing it plainly would just be a step back.

But I think we need to clarify on how we want to iterate over the codebase, because this is getting messy. I disagree about the fact that we cannot keep the codebase clean, it's just less "trivial" but it's an effort worth doing, as it also forces us to take into considerations other factors that may be outside the scope of our respective PRs.

recursion/src/verifier/batch_stark.rs
sai-deng: Added a TODO.

recursion/src/verifier/batch_stark.rs
sai-deng: Removed them.

Nashtare:
I don't want to be blocking your work for too long if that conversation goes on, so I won't insist further, but I strongly believe this is a discussion worth having and sooner than later.

recursion/src/verifier/batch_stark.rs
```diff
@@ -0,0 +1,561 @@
...
+        let flattened = ProofTargets {
+            commitments_targets: CommitmentTargets {
+                trace_targets,
+                quotient_chunks_targets,
+                random_commit: None,
+                _phantom: PhantomData,
+            },
+            opened_values_targets,
+            opening_proof: OpeningProof::new(circuit, &input.opening_proof),
+            degree_bits: 0,
```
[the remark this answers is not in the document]
sai-deng: Added comments.

recursion/src/verifier/stark.rs
```diff
@@ -188,15 +187,19 @@ where
...
-            zip_eq(
-                randomized_quotient_chunks_domains.iter(),
-                opened_quotient_chunks_targets,
-                VerificationError::InvalidProofShape(
-                    "Randomized quotient chunks length mismatch".to_string(),
-                ),
-            )?
-            .map(|(domain, values)| (*domain, vec![(zeta, values.clone())]))
-            .collect_vec(),
+            {
```
4l0n50: this is the same but, in my opinion, more verbose.

recursion/src/verifier/batch_stark.rs
```diff
@@ -0,0 +1,550 @@
...
+            Self::Witness(a) => a.eval(builder),
+            Self::Const(a) => a.eval(builder),
+            Self::Public(a) => a.eval(builder),
+            Self::Add(a) => a.eval(builder),
+            Self::Mul(a) => a.eval(builder),
+        }
+    }
+}
+
+/// Build and attach a recursive verifier circuit for a circuit-prover BatchStarkProof.
```
Nashtare:
```suggestion
/// Build and attach a recursive verifier circuit for a circuit-prover [`BatchStarkProof`].
```

Nashtare (approved):

PR #163, opened 2025-11-04
recursion tests: add `test_mul_verifier_circuit`

tcoratger:
@LindaGuiga Related to this comment https://github.com/Plonky3/Plonky3-recursion/pull/161#issuecomment-3484789967


LindaGuiga:
Thank you! But since we're bringing the test back at a stage where we already have RecursivePcs for TwoAdicFriPcs, maybe we can use that instead of the TrivialPcs? (It would also reduce the amount of code quite significantly)

sai-deng (approved):

PR #166, opened 2025-11-05
circuit prover: simplify w handling in batch stark prover

tcoratger:
Let me know if this makes sense but I guess that:
- `ExtractBinomialW` is not needed since we know at compile time if we have a base field or an extension field and what `W` is.
- I know that having a function for base field and one for extension could be a bit of code duplication but this is something that we are used to doing both in whir and Plonky3 for very low level optimizations so I say to myself that it's not so bad to do it here maybe, let me know :)


Nashtare (commented):
Overall this looks good, even though I'm not strongly opinionated towards this approach or the existing one, so I'll leave it to the others to see and decide.

FYI your tests are failing because you forgot to update the `mmcs_verify` example, should be using the new `prove_all_tables_extension` now. I've pushed a commit to fix it.

PR #167, opened 2025-11-06
feat: allow dynamic dispatch of non-primitive tables

Nashtare:
Complements #150 in allowing true dynamic handling of non-primitive operations at runtime, as opposed to the current prover which still _hardcodes_ tables to be proven like MMCS.

List of changes:

- Keep primitives hardcoded, and allow non-primitive ops to be added at runtime via prover plugin
- Introduce a `TableProver` trait with degree-specific methods: this one is a bit cumbersome, because we have to a distinction between degrees because of type-safety. I added a macro for easier handling so that we only need to specify the base case without worrying about extension ones.

### DIFF
--- a/circuit-prover/src/batch_stark_prover.rs
+++ b/circuit-prover/src/batch_stark_prover.rs
@@ -8,7 +8,8 @@ use alloc::{format, vec};
 
 use p3_air::{Air, BaseAir};
 use p3_batch_stark::{BatchProof, StarkGenericConfig, StarkInstance, Val};
-use p3_circuit::tables::Traces;
+use p3_circuit::ops::MmcsVerifyConfig;
+use p3_circuit::tables::{MmcsTrace, Traces};
 use p3_field::extension::BinomialExtensionField;
 use p3_field::{BasedVectorSpace, Field, PrimeField};
 use p3_matrix::dense::RowMajorMatrix;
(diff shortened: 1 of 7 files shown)


circuit/src/builder/circuit_builder.rs
```diff
@@ -67,6 +67,22 @@ where
...
+    pub fn enable_hash_squeeze(&mut self) {
+        self.config.enable_hash_squeeze();
+    }
+
+    /// Enables hash operations.
+    pub fn enable_hash(&mut self, reset: bool) {
+        self.enable_hash_absorb(reset);
+        self.enable_hash_squeeze();
+    }
+
```
Nashtare: Technically unrelated but I noticed these missing. I haven't kept track of the latest discussions around hashing handling so can remove these changes if needed.

circuit-prover/src/batch_stark_prover.rs
```diff
@@ -47,60 +55,365 @@ impl Default for TablePacking {
...
+    /// The populated trace matrix for this table.
+    pub trace: RowMajorMatrix<Val<SC>>,
+    /// Public values exposed by this table.
+    pub public_values: Vec<Val<SC>>,
+    /// Number of rows produced for this table.
+    pub rows: usize,
+}
+
+#[inline(always)]
+pub(crate) unsafe fn transmute_traces<FromEF, ToEF>(t: &Traces<FromEF>) -> &Traces<ToEF> {
```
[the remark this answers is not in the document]
Nashtare: I can add it.

circuit-prover/src/batch_stark_prover.rs
```diff
@@ -47,27 +55,342 @@ impl Default for TablePacking {
...
+    pub rows: usize,
+}
+
+#[inline(always)]
+/// # Safety
+///
+/// Caller must ensure that both `Traces<FromEF>` and `Traces<ToEF>` share an
+/// identical in-memory representation.
+pub(crate) unsafe fn transmute_traces<FromEF, ToEF>(t: &Traces<FromEF>) -> &Traces<ToEF> {
+    unsafe { &*(t as *const _ as *const Traces<ToEF>) }
```
sai-deng: I am wondering if we can avoid the unsafe path? If not, should we at least add
 debug_assert_eq! for size_of::<FromEF>(), size_of::<ToEF>(),
 and align_of::<FromEF>(), align_of::<ToEF>()
?

Nashtare: I don't think we can remove the unsafe, as the layouts are fundamentally different. But for sure I can add the debug assertions

sai-deng (approved):

circuit-prover/src/batch_stark_prover.rs
```diff
@@ -269,75 +683,142 @@ where
```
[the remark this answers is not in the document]
Nashtare: Done and added a test. Ideally we'd have some unified config between Circuit builder / Prover so that there is no need to re-do non-prim op registration and that configs remain consistent.

PR #169, opened 2025-11-06
Add witness hints fillers

4l0n50:
Currently, witness hints can be declared using `alloc_witness_hints`, but there is no defined mechanism for how these values are populated during circuit evaluation. This makes it unclear how unconstrained data should be assigned within the witness table.

This PR introduces a new primitive operation, `Op::Unconstrained`, which assigns the values of witness hints. During evaluation time, these values are populated according to the behavior defined by the new WitnessHintFiller trait. Implementors of this trait define the logic for deriving hint values, potentially using existing wire values already computed in the circuit.

### DIFF
--- a/circuit/src/circuit.rs
+++ b/circuit/src/circuit.rs
@@ -136,9 +136,6 @@ impl<F: Field> Circuit<F> {
                     max_idx = max_idx.max(a.0).max(b.0).max(out.0);
                 }
                 Op::Unconstrained { outputs, .. } => {
-                    let table_idx = PrimitiveOpType::Witness as usize;
-                    preprocessed[table_idx]
-                        .extend(outputs.iter().map(|output| F::from_u32(output.0)));
                     max_idx = iter::once(max_idx)
                         .chain(outputs.iter().map(|&output| output.0))
                         .max()


circuit/src/builder/compiler/expression_lowerer.rs
```diff
@@ -179,18 +189,57 @@ where
             }
         }
 
-        // Pass C: emit arithmetic ops in creation order; tie outputs to class slot if connected
+        // Pass C: collect witness hints and emit `Unconstrained` operatios
```
Nashtare:
```suggestion
        // Pass C: collect witness hints and emit `Unconstrained` operations
```

circuit/src/builder/circuit_builder.rs
```diff
@@ -700,6 +713,68 @@ mod tests {
         assert_eq!(circuit.primitive_ops.len(), 2);
     }
 
+    #[derive(Debug, Clone)]
+    struct ConstantHint<const C: usize> {
+        inputs: Vec<ExprId>,
+    }
```
Nashtare: What do the `inputs` represent here? More generally why would we need a `ConstantHint`? Constants are preprocessed so fine to have them via the `Op::Const` type.

circuit/src/builder/compiler/expression_lowerer.rs
```diff
@@ -179,18 +189,58 @@ where
...
```
Nashtare: Question: How does ordering work if, for instance, I have some `WitnessHint` whose inputs depend on an arithmetic op, as these are handled _after_ the witness pass? Before this couldn't happen because we were handling all primitive ops (apart from Constant / PI) together, but now I think we may throw a `MissingExprMapping` issue in some cases?

circuit/src/builder/errors.rs
```diff
@@ -23,6 +23,10 @@ pub enum CircuitBuilderError {
         got: usize,
     },
 
+    /// Unconstrained op received an unexpected number of input expressions.
+    #[error("Expects exactly {expected} witness expressions, got {got}")]
+    UnconstrainedeOpArity { expected: String, got: usize },
+
```
Nashtare: Also this is actually never used?

circuit/src/tables/runner.rs
```diff
@@ -412,6 +438,148 @@ mod tests {
         assert_eq!(traces.add_trace.result_index, vec![WitnessId(4)]);
     }
 
+    #[derive(Debug, Clone)]
+    /// The hint deined by x in an equation a*x - b = 0
+    struct X {
```
Nashtare:
```suggestion
    struct XHint {
```
perhaps?

Nashtare (commented):
Thanks Alonso!

As a general comment, I feel like `Unconstrained` is a bit misleading, as we may still apply constraints on these values (like in `hash_squeeze` for instance), so perhaps it'd be worth thinking of another name, maybe something like `Op::Private` or anything that reflects the underlying logic?

circuit/src/builder/compiler/expression_lowerer.rs
4l0n50: You're right. I need to move the push of `Op::Unconstrained` after Pass D

4l0n50: Now should be fixed

circuit/src/builder/expression_builder.rs
```diff
@@ -100,6 +108,7 @@ where
 
     /// Adds a witness hint to the graph.
     /// It will allocate a `WitnessId` during lowering, with no primitive op.
+    /// TODO: Make this function private.
```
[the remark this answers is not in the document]
Nashtare: but it's not even used for now, is it?

circuit/src/utils.rs
```diff
@@ -126,34 +129,77 @@ pub fn reconstruct_index_from_bits<F: Field>(
...
-    let mut bits = Vec::with_capacity(n_bits);
-
     // Create bit witness variables
-    for _ in 0..n_bits {
-        let bit = builder.add_public_input(); // TODO: Should be witness
-        builder.assert_bool(bit);
-        bits.push(bit);
-    }
+    let binary_decomposition_hint = BinaryDecompositionHint::new(x, n_bits)?;
+    let bits = builder.alloc_witness_hints(binary_decomposition_hint, "decompose_to_bits");
```
sai-deng: Should we add back builder.assert_bool(bit);?

circuit/src/tables/runner.rs
```diff
@@ -412,6 +438,148 @@ mod tests {
...
+        assert!(traces.public_trace.values.is_empty());
+
+        // Should have one mul operation: 37 * x
+        assert_eq!(traces.mul_trace.lhs_values.len(), 1);
+
+        // Encoded subtraction lands in the add table (result + rhs = lhs).
+        assert_eq!(traces.add_trace.lhs_values.len(), 1);
+        assert_eq!(traces.add_trace.lhs_index, vec![WitnessId(2)]);
+        assert_eq!(traces.add_trace.rhs_index, vec![WitnessId(0)]);
+        assert_eq!(traces.add_trace.result_index, vec![WitnessId(4)]);
```
LindaGuiga: Can we add an assert that the `x` value is 3?

circuit/src/utils.rs
4l0n50: ops, you're right!

circuit/src/tables/mmcs.rs
```diff
@@ -3,6 +3,7 @@ use alloc::vec::Vec;
 use alloc::{format, vec};
 use core::fmt::Debug;
 use core::iter;
+use core::result::Result;
```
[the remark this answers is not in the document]
4l0n50: I can remove it without problems, even though clippy was not complaining 😕

recursion/src/traits/challenger.rs
```diff
@@ -82,16 +82,19 @@ pub trait RecursiveChallenger<F: Field> {
...
-    ) -> Vec<Target> {
+    ) -> Vec<Target>
+    where
+        F: ExtensionField<BF>,
+    {
         let x = self.sample(circuit);
 
         // Decompose to bits (adds public inputs for each bit and verifies they reconstruct x)
-        let bits = decompose_to_bits(circuit, x, total_num_bits);
+        let bits = decompose_to_bits(circuit, x, total_num_bits).unwrap();
```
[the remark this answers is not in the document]
4l0n50: ok, but this will need to change the signature of many other functions

Nashtare:
FYI @4l0n50 I've pulled in latest main to get CI to go ✅ 
Also noticed you reintroduced the `prover` module (with `MultiTableProver`) in this PR, I assume it's a mistake? It's been removed in #170

circuit/src/tables/runner.rs
```diff
@@ -412,6 +438,148 @@ mod tests {
...
+
+        let mul_result = builder.mul(c37, x);
+        let sub_result = builder.sub(mul_result, c111);
+        builder.assert_zero(sub_result);
+
+        let circuit = builder.build().unwrap();
+        println!("=== CIRCUIT PRIMITIVE OPERATIONS ===");
+        for (i, prim) in circuit.primitive_ops.iter().enumerate() {
+            println!("{i}: {prim:?}");
+        }
```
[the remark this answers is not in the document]
Nashtare: sorry I had missed your comment!

We can probably remove the former one and keep this one, the goal was to highlight basic operations logic and yours does this + includes hints.

However my comment still stands, now that we have the journal for allocations that can be dumped we don't need those prints.

circuit/src/utils.rs
```diff
@@ -126,25 +129,72 @@ pub fn reconstruct_index_from_bits<F: Field>(
     acc
 }
 
+#[derive(Debug, Clone)]
+/// Given a field element as input, decompose it into its little-endian bits and
+/// fill witness hints with the binary decomposition.
+///
+/// For a given input `input`, fills `n_bits` witness hints with `b_i`
+/// such that that:
```
Nashtare:
```suggestion
/// such that:
```

circuit/src/utils.rs
```diff
@@ -126,25 +129,72 @@ pub fn reconstruct_index_from_bits<F: Field>(
...
+}
+
+impl<BF: PrimeField64> BinaryDecompositionHint<BF> {
+    pub fn new(input: ExprId, n_bits: usize) -> Result<Self, CircuitError> {
+        if n_bits > 64 {
+            return Err(CircuitError::UnconstrainedOpInputLengthMismatch {
+                expected: 64,
+                got: n_bits,
+            });
+        }
```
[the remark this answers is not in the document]
4l0n50: we can change the error to have a string on expected, or another field indicating wether the comparison operator is `>=`, `==`, etc.

recursion/src/traits/challenger.rs
```diff
@@ -82,18 +82,21 @@ pub trait RecursiveChallenger<F: Field> {
...
         total_num_bits: usize,
         num_bits: usize,
-    ) -> Vec<Target> {
+    ) -> Result<Vec<Target>, CircuitError>
+    where
+        F: ExtensionField<BF>,
+    {
         let x = self.sample(circuit);
 
         // Decompose to bits (adds public inputs for each bit and verifies they reconstruct x)
```
sai-deng: update comments (remove "public inputs")

recursion/src/traits/challenger.rs
sai-deng: update comments (remove "public inputs")

circuit/src/utils.rs
```diff
@@ -126,25 +130,73 @@ pub fn reconstruct_index_from_bits<F: Field>(
...
+struct BinaryDecompositionHint<BF: PrimeField64> {
+    inputs: Vec<ExprId>,
+    n_bits: usize,
+    _phantom: PhantomData<BF>,
+}
+
+impl<BF: PrimeField64> BinaryDecompositionHint<BF> {
+    pub fn new(input: ExprId, n_bits: usize) -> Result<Self, CircuitError> {
+        if n_bits > 64 {
+            return Err(CircuitError::UnconstrainedOpInputLengthMismatch {
```
sai-deng: Maybe add a new error type for BinaryDecompositionTooManyBits

sai-deng (approved):

circuit/src/tables/mod.rs
```diff
@@ -35,3 +35,69 @@ pub struct Traces<F> {
...
+
+        tracing::debug!("\n=== CONST TRACE ===");
+        for (i, (idx, val)) in self
+            .const_trace
+            .index
+            .iter()
+            .zip(self.const_trace.values.iter())
+            .enumerate()
+        {
+            tracing::debug!("Row {i}: WitnessId({idx}) = {val:?}");
```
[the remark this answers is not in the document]
4l0n50: why? that would mean mul trace prints `MulIdx`?

circuit/src/utils.rs
4l0n50: added

PR #170, opened 2025-11-06
Remove MultiTableProver

sai-deng:


circuit-prover/src/batch_stark_prover.rs
```diff
@@ -16,7 +16,40 @@ use tracing::instrument;
 use crate::air::{AddAir, ConstAir, MulAir, PublicAir, WitnessAir};
 use crate::config::StarkField;
 use crate::field_params::ExtractBinomialW;
-use crate::prover::TablePacking;
+
+// Packing configuration for Add/Mul tables.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
Nashtare: nit: could we keep the old description before the move to clarify a bit its meaning? `TablePacking` may be a bit opaque to newcomers otherwise without diving in the code

PR #172, opened 2025-11-11
circuit prover: refactor and documentation for `MulAir`

tcoratger:
Similar to https://github.com/Plonky3/Plonky3-recursion/pull/164 but for `MulAir`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,24 +26,24 @@ categories = ["cryptography::cryptocurrencies"]
 [workspace.dependencies]
 
 # Plonky3 dependencies - shared across all crates
-p3-air = { git = "https://github.com/Plonky3/Plonky3" }
-p3-baby-bear = { git = "https://github.com/Plonky3/Plonky3" }
-p3-batch-stark = { git = "https://github.com/Plonky3/Plonky3" }
-p3-challenger = { git = "https://github.com/Plonky3/Plonky3" }
-p3-circle = { git = "https://github.com/Plonky3/Plonky3" }
-p3-commit = { git = "https://github.com/Plonky3/Plonky3" }
-p3-dft = { git = "https://github.com/Plonky3/Plonky3" }
-p3-field = { git = "https://github.com/Plonky3/Plonky3" }
-p3-fri = { git = "https://github.com/Plonky3/Plonky3" }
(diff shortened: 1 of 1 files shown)


circuit-prover/src/air/mul_air.rs
```diff
@@ -45,34 +88,45 @@ use p3_matrix::Matrix;
...
 }
 
 impl<F: Field + PrimeCharacteristicRing, const D: usize> MulAir<F, D> {
-    /// Constructor for base-field or non-binomial cases (`D == 1`).
+    /// Construct a new `MulAir` for base-field operations (D=1).
+    ///
+    /// Panics if `lanes == 0` or `D > 1`.
     pub const fn new(num_ops: usize, lanes: usize) -> Self {
         assert!(lanes > 0, "lane count must be non-zero");
+        assert!(D == 1, "Use new_binomial for D > 1");
```
tcoratger: Here I think that we need to check that `D == 1` and panic if this is not the case.

circuit-prover/src/air/mul_air.rs
```diff
@@ -94,64 +149,143 @@ impl<F: Field + PrimeCharacteristicRing, const D: usize> MulAir<F, D> {
...
+        for (op_idx, (((((lhs_val, lhs_idx), rhs_val), rhs_idx), res_val), res_idx)) in trace
+            .lhs_values
+            .iter()
+            .zip(trace.lhs_index.iter())
+            .zip(trace.rhs_values.iter())
+            .zip(trace.rhs_index.iter())
+            .zip(trace.result_values.iter())
+            .zip(trace.result_index.iter())
+            .enumerate()
+        {
```
tcoratger: With this we can avoid the double loop I think

circuit-prover/src/air/mul_air.rs
```diff
@@ -167,57 +301,140 @@ where
...
+                            //
+                            // So our full term (term * x^k) becomes:
+                            //   term * (W * x^{k-D})
+                            //
+                            // This is a new term of degree (k-D), with a new
+                            // coefficient of (term * W).
+                            //
+                            // Math: acc[k-D] = acc[k-D] + (term * W)
+                            acc[k - D] += w.clone() * term;
+                        }
```
tcoratger: Here I've added explanations because this is a bit tricky to understand

PR #174, opened 2025-11-13
Add preprocessing to `verify_circuit`

LindaGuiga:
This PR aims at adding preprocessing to `verify_circuit` (but not the batch version, as it is not yet supported on the Pllonky3 side).
It also adds preprocessing to `test_mul_verifier` so the recursive preprocessing is tested.

Finally, it adds a fifth step in circuit building (in circuit_builder.rs) to generate preprocessed values (not matrices as the width depends on the number of lanes for Add and Mul) for the primitive operations. I have not yet integrated the preprocessed values to the tables, to keep this PR limited and because we cannot yet leverage the preprocessing in batch proving/verifying.

### DIFF
--- a/circuit/src/circuit.rs
+++ b/circuit/src/circuit.rs
@@ -8,7 +8,7 @@ use p3_field::Field;
 use strum::EnumCount;
 
 use crate::op::{NonPrimitiveOpConfig, NonPrimitiveOpType, Op, PrimitiveOpType};
-use crate::tables::CircuitRunner;
+use crate::tables::{CircuitRunner, TraceGeneratorFn};
 use crate::types::{ExprId, WitnessId};
 
 /// Trait encapsulating the required field operations for circuits
@@ -60,6 +60,8 @@ pub struct Circuit<F> {
     /// Enabled non-primitive operation types with their respective configuration
     pub enabled_ops: HashMap<NonPrimitiveOpType, NonPrimitiveOpConfig>,
     pub expr_to_widx: HashMap<ExprId, WitnessId>,
+    /// Registered non-primitive trace generators.
+    pub non_primitive_trace_generators: HashMap<NonPrimitiveOpType, TraceGeneratorFn<F>>,
 }
 
 impl<F: Field + Clone> Clone for Circuit<F> {
@@ -72,6 +74,7 @@ impl<F: Field + Clone> Clone for Circuit<F> {
             public_flat_len: self.public_flat_len,
             enabled_ops: self.enabled_ops.clone(),
             expr_to_widx: self.expr_to_widx.clone(),
+            non_primitive_trace_generators: self.non_primitive_trace_generators.clone(),
         }
     }
 }
@@ -86,6 +89,7 @@ impl<F: Field> Circuit<F> {
             public_flat_len: 0,
             enabled_ops: HashMap::new(),
             expr_to_widx,
+            non_primitive_trace_generators: HashMap::new(),
         }
     }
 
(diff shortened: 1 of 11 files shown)


circuit-prover/examples/fibonacci.rs
```diff
@@ -54,7 +54,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     builder.dump_allocation_log();
 
-    let circuit = builder.build()?;
+    let (circuit, _) = builder.build()?;
```
sai-deng: I think it might be nicer to store preprocessed cols in Circuit instead of returning them separately.

circuit/src/circuit.rs
```diff
@@ -86,6 +88,60 @@ impl<F> Circuit<F> {
...
+                    let table_idx = PrimitiveOpType::Mul as usize;
+                    preprocessed[table_idx].extend(&[
+                        F::from_u32(a.0),
+                        F::from_u32(b.0),
+                        F::from_u32(out.0),
+                    ]);
+                    max_idx = max_idx.max(a.0).max(b.0).max(out.0);
+                }
+                Op::NonPrimitiveOpWithExecutor { .. } => panic!(
+                    "preprocessed values are not yet implemented for non primitivie operations."
```
sai-deng: typo: primitive

circuit/src/op.rs
```diff
@@ -34,7 +35,7 @@ use crate::types::{NonPrimitiveOpId, WitnessId};
 ///
 /// They can be user-defined and selected at runtime, have private data that does not appear
 /// in the central Witness bus, and are subject to their own optimization passes.
-#[derive(Debug)]
+#[derive(Debug, EnumCount)]
```
sai-deng: EnumCount derive is not needed on Op?

circuit/src/tables/mod.rs
```diff
@@ -1,6 +1,6 @@
 //! Execution trace tables for zkVM circuit operations.
 
-mod add;
+pub mod add;
```
Nashtare: nit: why the change?

recursion/tests/mul_air.rs
```diff
@@ -40,10 +40,11 @@ type MyPcs = TwoAdicFriPcs<F, Dft, ValMmcs, ChallengeMmcs>;
 type MyConfig = StarkConfig<MyPcs, Challenge, Challenger>;
 
 /// Number of repetitions of the multiplication constraint (must be < 255 to fit in u8)
-const REPETITIONS: usize = 20;
+const REPETITIONS: usize = 1;
```
Nashtare: nit: revert?

circuit-prover/examples/fibonacci.rs
LindaGuiga: Well, I didn't want to actually store them in Circuit so people can write them in (and read them from) a file instead. I can't directly do that because of no-std (and we don't know the system will be using). 

Also, what we actually want to store later is the commits to the values, but they depend on the config and `num_lanes`. So we might want to store commits for multiple configs. I was afraid it might become heavy/messy for the Circuit if we store them there.

sai-deng: Ideally, we should not return `Vec<Vec<F>>` for preprocessed cols. But this can be done in a followup PR.

sai-deng: Actually, since only the index columns are preprocessed columns, I think we can have the prover handle them directly (same as CTL). There’s no need to return anything from circuit.build().

PR #179, opened 2025-11-19
feat: integrate Poseidon2

Nashtare:
Builds over #131 and integrates Poseidon2 inside the circuit.
Includes a `hash_chain` example to highlight use of `HashAbsorb` / `HashSqueeze`.

I should be opening soon another PR to integrate hashing directly in MMCS (via a `HashCompress` instruction most likely) reusing the Poseidon2 table, so that we can have MMCS / FRI leverage the new hasher.

Note that in a few places, we have to manually hardcode the configurations to be supported (field / ext. degree / width) because of const generics issue. For now I've added only support for Koala / Baby bear fields but we could add Goldilocks support too without too much trouble I'd assume.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           command: install
           args: mdbook --version 0.4.52
-
+      
       - name: Install katex preprocessor
         uses: actions-rs/cargo@v1
         with:
--- a/circuit-prover/src/air/const_air.rs
+++ b/circuit-prover/src/air/const_air.rs
@@ -1,11 +1,18 @@
 //! [`ConstAir`] stores constants either in the base field or the extension field (of extension degree `D`).
 //!
-//!  # Columns
+//! # Column layout
 //!
-//! The AIR has `D + 1` columns:
+//! The AIR is generic over an extension degree `D`.
+//! For each constant entry, we allocate `D + 1` base-field columns.
 //!
-//! - 1 column for the index of the constant within the witness table,
-//! - `D` columns for the constant value.
+//! - `D` columns for the constant value (basis coefficients),
+//! - `1` column for the `index`: the witness-bus index of the constant.
+//!
+//! The layout for a single row is:
+//!
+//! ```text
+//!     [value[0], value[1], ..., value[D-1], index]
+//! ```
 //!
 //! # Constraints
 //!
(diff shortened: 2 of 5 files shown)


poseidon2-circuit-air/src/air.rs
```diff
@@ -0,0 +1,622 @@
...
+            PARTIAL_ROUNDS,
+        >();
+
+        debug_assert_eq!(ncols, num_circuit_cols + p2_ncols);
+
+        let mut vec = vec![F::ZERO; n * ncols];
+
+        let circuit_trace_view = circuit_trace.as_view();
+
+        for ((row, left_part), right_part) in vec
```
4l0n50: Here we need to add TODO for removing copying the already computed `p2_trace`

4l0n50: Actually, this would be easy if [`p3_poseidon2_air::generate_trace_rows_for_perm`](https://github.com/Plonky3/Plonky3/blob/7656e166e044cb7955b783ce4108e87e1cecf47e/poseidon2-air/src/generation.rs#L134) becomes public. We would just need to declare the trace matrix as
```rust
let num_permutation_cols = p3_poseidon2_air::num_cols::<
            WIDTH,
            SBOX_DEGREE,
            SBOX_REGISTERS,
            HALF_FULL_ROUNDS,
            PARTIAL_ROUNDS,
        >();
let mut circuit_trace = vec![F::ZERO; n * (num_circuit_cols + num_permutation_cols)];
```
and write each poseidon2 permutation row, using `p3_poseidon2_air::generate_trace_rows_for_perm`, at the end of each row.

poseidon2-circuit-air/src/columns.rs
```diff
@@ -0,0 +1,79 @@
...
+/// memory lookup. Only used by compressions to send the output.
+#[repr(C)]
+pub struct Poseidon2CircuitCols<
+    T,
+    const WIDTH_EXT: usize,
+    const RATE_EXT: usize,
+    P: PermutationColumns<T>,
+> {
+    pub poseidon2: P,
+    pub is_sponge: T,
```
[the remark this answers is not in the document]
Nashtare: I've just copied Hamy's code and added what's left to make this PR consistent and mergeable (the other PR only adds code that isn't used by the rest of the system), but didn't apply any changes to the AIR itself.

4l0n50 (approved):
LGTM. We can add an issue for removing the extra copy in `generate_trace_rows` once this is approved.

Issue #162, opened 2025-10-30
Removal of MMCS Table


sai-deng:
I propose removing the MMCS table from the current design.

The purpose of the MMCS table is to hide certain private inputs. However, almost all its rows are already connected to the Poseidon table.

A better approach would be to use a single, modified Poseidon table for all operational logic in both the challenger and MMCS.
This new "Poseidon2" table will introduce additional columns, such as:
```
reset
is_output_row
has_private_inputs
input and output index columns
```

With this design, each row of the Poseidon2 table represents a generic operation of the form:
`PHash(input extension elements, [optional private input extension elements]) = output extension elements
`

For MMCS verification, the operation becomes a circuit, making calls like:
(a segment is the layer with leaves to the next layer with leaves)

```
PHash(leaves_seg0) = leaf_hash_0
PHash(leaves_seg1) = leaf_hash_1
...
PHash(leaf_hash_0, [Merkle path]) = intermediate_hash_0
PHash(intermediate_hash_0, leaf_hash_1, [Merkle path]) = intermediate_hash_1 or new_root (if this is the last segment)
...
Finally, the following assertion is made in the circuit:
assert_eq(root, new_root)
```

4l0n50:
> `PHash(leaves_seg0) = leaf_hash_0`

`leaves_seg0` would correspond to the openings of one of the committed matrices at that index, right?

> `PHash(leaf_hash_0, [Merkle path]) = intermediate_hash_0`

Wouldn't it look more like this?

```
select(direction_0, leaf_hash_0, sibling_0) = left_0
leaf_hash_0 + sibling_0 - left_0 = right_0
PHash(left_0, right_0) = state_1
...
PHash(left_n, right_n) = intermediate_hash_0
PHash(intermediate_hash, leaf_hash_1)
...
```

I think this would change the modifications you proposed to the Poseidon table — otherwise, you'd also need to include the directions as inputs to `PHash`.
In any case, I agree this approach would still be better than having two separate tables, since the MMCS table calls Poseidon at every row. That means we'd at least save the lookups.

sai-deng:
Yes, you are right. I forgot the directions.

4l0n50:
I was thinking we could also do the switch in this table, instead of delegating it to the mul, add tables. But I'm not sure if that would be better or not, so I'm ok with what you say

sai-deng:
[...]

PR #181, opened 2025-11-19
feat: plug Poseidon2 into MMCS

Nashtare:
Add a `NonPrimitiveOpType::HashCompress` operation to perform 2-to-1 compression to be used in MMCS.
Plug in Poseidon2 logic within `verify_mmcs` integration test.

Still requires to feed MMCS private data externally to the runner, this should be removed in a follow-up PR.
Remains to finish the FRI logic.
We should also be able to relatively easily remove the MMCS table in favor of an in-circuit one now that the hashing logic is plugged in, cf #162

### DIFF
--- a/circuit/src/ops/hash.rs
+++ b/circuit/src/ops/hash.rs
@@ -127,7 +127,7 @@ where
 
         let _ = self.push_non_primitive_op(
             NonPrimitiveOpType::HashCompress,
-            vec![left.to_vec(), right.to_vec()],
+            vec![left.to_vec(), right.to_vec(), outputs.to_vec()],
             "HashCompress",
         );
 


circuit/src/ops/hash.rs
```diff
@@ -92,6 +101,38 @@ where
...
+                    right.len()
+                ),
+                got: 0,
+            });
+        }
+
+        // Allocate output hints
+        let outputs = self.alloc_witness_hints_default_filler(left.len(), "hash_compress_output");
+
+        let _ = self.push_non_primitive_op(
```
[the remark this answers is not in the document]
Nashtare: I've added them to the `witness_exprs` of the non-prim op too

circuit/src/builder/compiler/non_primitive_lowerer.rs
```diff
@@ -197,6 +200,62 @@ impl<'a> NonPrimitiveLowerer<'a> {
                         op_id: *op_id,
                     });
                 }
+                NonPrimitiveOpType::HashCompress => {
```
sai-deng: (same as what I shared in the channel)
Currently we have `NonPrimitiveOpType::HashAbsorb/HashSqueeze/HashCompress`. But in my [design](https://github.com/Plonky3/Plonky3-recursion/issues/162), we do not need these ops. We only need one operation that is Perm().  Absorb/Squeeze/MMCS verify becomes circuits.
The OP:
Perm(in: Option<F[4]>, out: Option<F[4]>, new_start :bool, in_ctl: bool[4], out_ctl: bool)

sai-deng: This OP by default connects all outputs to next Perm()'s inputs.

PR #184, opened 2025-11-20
Compute hash squeeze hints

hratoanina:
This PR only refactors the HashSqueeze op and implements its filler without touching the recursion crate for now.

It's functional, but some tests are missing to open it fully.

### DIFF
--- a/circuit/src/ops/hash.rs
+++ b/circuit/src/ops/hash.rs
@@ -281,7 +281,6 @@ mod tests {
             .chunks(4)
             .zip(intermediate_digest)
         {
-            // The first two witnesses are constants 0 and 1.
             let expected =
                 circuit_builder.add_const(CF::from_basis_coefficients_slice(chunk).unwrap());
             circuit_builder.connect(expected, value);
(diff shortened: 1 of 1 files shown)


circuit/src/op.rs
```diff
@@ -445,7 +461,11 @@ pub trait WitnessHintsFiller<F>: Debug + WitnessFillerClone<F> {
     /// Compute the output given the inputs
     /// # Arguments
     /// * `inputs` - Input witness
-    fn compute_outputs(&self, inputs_val: Vec<F>) -> Result<Vec<F>, CircuitError>;
+    fn compute_outputs(
+        &self,
+        inputs_val: Vec<F>,
```
4l0n50: If you add the context you can get rid of `inputs_val`, because they can be derived from the context

sai-deng:
Should we close this PR and implement it as a circuit?

4l0n50:
> Should we close this PR and implement it as a circuit?

Unless I'm missing something I would say no. Right now is functional (maybe still fixing the bug Hamy mentioned?) and the interface should be the same even after using the new squeeze/absorb/mmcs table. That means that after merging this PR we can continue adding hashing/mmcs to fri. I just took a very quick look to your PR with the new table and I think the change would be just pushing the new `NonPrimitiveOpType::Perm` non-primitive operation, instead of pushing `NonPrimitiveOpType::HashAbsorb, HashSqueeze, MmmcsVerify`. 
Do you disagree? because I was planning on finishing this PR on Monday.

PR #189, opened 2025-11-21
tables constants: add unit tests

tcoratger:


circuit/src/tables/constant.rs
```diff
@@ -45,3 +45,124 @@ impl<'a, F: Clone> ConstTraceBuilder<'a, F> {
...
+        let trace = builder.build().expect("Failed to build trace");
+
+        assert_eq!(trace.index.len(), 2);
+        assert_eq!(trace.values.len(), 2);
+
+        assert_eq!(trace.index[0], out);
+        assert_eq!(trace.values[0], val1);
+        assert_eq!(trace.index[1], out);
+        assert_eq!(trace.values[1], val2);
+    }
```
tcoratger: I'm very surprised this works this way, shouldn't we error out with `CircuitError::WitnessConflict` in this case?

PR #190, opened 2025-11-21
tables: small touchups

tcoratger:


Nashtare (approved):
LGTM. May be worth adding it to lint rules?

PR #192, opened 2025-11-24
bug: fix ci parsing

Nashtare:
Seems that there are some parsing issues lately.
See for instance https://github.com/Plonky3/Plonky3-recursion/actions/runs/19602397346/job/56135792412.


Nashtare:
Actually this is unrelated to our `.yml` files but this is an error upstream in Pulumi.

Issue #182, opened 2025-11-20
perf: remove needless copy of Plonky3 Poseidon2 trace


Nashtare:
if Plonky3 exposed `generate_trace_rows_for_perm`, we could remove an entire vector allocation + cloning when combining the Poseidon2 subtrace with the extra circuit / recursion table in `poseidon2-circuit-air/src/air.rs`.

PR #194, opened 2025-11-24
perf: remove needless copy of Plonky3 Poseidon2 trace

tcoratger:
Should close https://github.com/Plonky3/Plonky3-recursion/issues/182

That is possible after https://github.com/Plonky3/Plonky3/pull/1159

### DIFF
--- a/recursion/src/generation.rs
+++ b/recursion/src/generation.rs
@@ -3,8 +3,6 @@ use alloc::vec::Vec;
 
 use itertools::zip_eq;
 use p3_air::Air;
-use p3_batch_stark::BatchProof;
-use p3_batch_stark::config::observe_instance_binding;
 use p3_batch_stark::config::observe_instance_binding;
 use p3_batch_stark::{BatchProof, CommonData};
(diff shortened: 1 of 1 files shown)


poseidon2-circuit-air/src/air.rs
```diff
@@ -101,13 +102,30 @@ impl<
...
+            SBOX_REGISTERS,
+            HALF_FULL_ROUNDS,
+            PARTIAL_ROUNDS,
+        >();
 
-        let mut state = [F::ZERO; WIDTH];
+        let ncols = self.width();
+
+        // We allocate the final vector immediately with uninitialized memory.
+        let mut trace_vec: Vec<F> = Vec::with_capacity((n * ncols) << extra_capacity_bits);
```
4l0n50: I think the `<< extra_capacity_bits` should only enlarge the p2 cols.

4l0n50 (commented):
I opened a similar PR but with a less accurate name. The p2 trace generation was more naive, so I'm closing it in favor of this one. Thanks!

Nashtare (approved):
LGTM, thanks!

PR #196, opened 2025-11-25
Add preprocessing to primitive operations

LindaGuiga:
This PR aims at adding preprocessing to primitive operations. It adds a new `common.rs` file in the crate `circuit-prover` that handles common data.
Note that in order for the AIRs to implement `PairBuilder` (and more specifically `fn preprocessed_trace(&self)`, I added a `preprocessed: Vec<F>` field to the AIRs . We call `get_airs_and_degrees_with_prep` to get AIRs that are properly setup for preprocessing, along with their degrees (the latter is useful to then call `from_airs_and_degrees` to get the common data)

### DIFF
--- a/circuit-prover/src/air/add_air.rs
+++ b/circuit-prover/src/air/add_air.rs
@@ -32,17 +32,20 @@
 //! grouped as:
 //!
 //! - `D` columns for the left operand (lhs) basis coefficients,
-//! - `1` column for `index_left`: the witness-bus index of the lhs operand,
 //! - `D` columns for the right operand (rhs) basis coefficients,
-//! - `1` column for `index_right`: the witness-bus index of the rhs operand,
 //! - `D` columns for the result operand basis coefficients,
-//! - `1` column for `index_output`: the witness-bus index of the result.
 //!
 //! In other words, for a single lane the layout is:
 //!
 //! ```text
-//!     [lhs[0..D), lhs_index, rhs[0..D), rhs_index, result[0..D), result_index]
+//!     [lhs[0..D), rhs[0..D), result[0..D)]
 //! ```
+//!
+//! We also allocate 3 preprocessed base-field columns:
+//!
+//! - 1 column for the lhs operand index within the `Witness` table,
+//! - 1 column for the rhs operand index,
+//! - 1 column for the result operand index,
 //!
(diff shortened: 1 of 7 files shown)


circuit-prover/src/air/add_air.rs
```diff
@@ -102,6 +104,10 @@ pub struct AddAir<F, const D: usize = 1> {
     ///
     /// The last row is padded if the number of operations is not a multiple of this value.
     pub lanes: usize,
+    /// Preprocessed values. They are used for generating the common data,
+    /// as well as by the prover, to compute the constraint polynomial.
+    /// Preprocessed values correspond to the indices of the inputs and outpurs withing the `Witness`.
```
Nashtare:
```suggestion
    /// Preprocessed values correspond to the indices of the inputs and outputs within the `Witness`.
```

circuit-prover/src/air/mul_air.rs
```diff
@@ -122,6 +123,10 @@ pub struct MulAir<F, const D: usize = 1> {
     /// - `Some(W)` if `D > 1`,
     /// - `None` if `D == 1`.
     pub w_binomial: Option<F>,
+    /// Preprocessed values. They are used for generating the common data,
+    /// as well as by the prover, to compute the constraint polynomial.
+    /// Preprocessed values correspond to the indices of the inputs and outpurs withing the `Witness`.
```
Nashtare:
```suggestion
    /// Preprocessed values correspond to the indices of the inputs and outputs within the `Witness`.
```

circuit-prover/src/common.rs
```diff
@@ -0,0 +1,111 @@
...
+use crate::field_params::ExtractBinomialW;
+use alloc::vec::Vec;
+use p3_circuit::Circuit;
+use p3_circuit::CircuitError;
+use p3_circuit::op::PrimitiveOpType;
+use p3_field::ExtensionField;
+use p3_uni_stark::StarkGenericConfig;
+use p3_uni_stark::Val;
+use p3_util::log2_ceil_usize;
+use strum::EnumCount;
```
Nashtare: not needed? Also it's import in the config file

LindaGuiga: `EnumCount` is needed for `PrimitiveOpType::COUNT`. Compiling fails if I remove it

PR #197, opened 2025-11-26
rename: `Expr::Witness` to `Expr::Hint`

Nashtare:
No functional change, just renaming the `Expr::Witness` to `Expr::Hint` as it may carry the underlying intended behavior better and removes any possible confusion with `PrimitiveOpType::Witness` which refers to the `Witness` Table, and not the witness hints.


LindaGuiga (approved):
LGTM, thank you!

PR #202, opened 2025-12-01
refactor: make Poseidon2 generic again over some chosen config

Nashtare:
Addresses concerns expressed in #187 that removed the modularity w.r.t to some chosen config (base field, permutation width, ... for Poseidon2).

The approach is basically defining two types: a const-generic one to be used by the AIR / prover and a dynamic one for trait-objects (as they can't deal with const-generic), although we may as well get rid of the const-generic altogether and solely rely on the dynamic version for clarity.

I haven't checked godbolt but given that the associated const generics are, well const, I'd assume all instantiations like 

```rust
let width_ext = Config::WIDTH_EXT;
let mut in_ctl = vec![false; width_ext];
```
        
to be essentially free

### DIFF
--- a/circuit/src/ops/poseidon2_perm.rs
+++ b/circuit/src/ops/poseidon2_perm.rs
@@ -428,11 +428,9 @@ impl Poseidon2PermExecutor {
                         resolved[2] = Some(prev[0]);
                         resolved[3] = Some(prev[1]);
                     }
-                } else {
-                    if prev.len() >= digest_ext {
-                        resolved[0] = Some(prev[0]);
-                        resolved[1] = Some(prev[1]);
-                    }
+                } else if prev.len() >= digest_ext {
+                    resolved[0] = Some(prev[0]);
+                    resolved[1] = Some(prev[1]);
                 }
             }
         }
--- a/circuit/src/tables/poseidon2.rs
+++ b/circuit/src/tables/poseidon2.rs
@@ -537,26 +537,10 @@ impl<F> Poseidon2TraceDyn<F> {
                     mmcs_bit: row.mmcs_bit,
                     mmcs_index_sum: row.mmcs_index_sum.clone(),
                     input_values: row.input_values.clone(),
-                    in_ctl: row
-                        .in_ctl
-                        .clone()
-                        .try_into()
-                        .map_err(|_| CircuitError::InvalidTraceData)?,
-                    input_indices: row
-                        .input_indices
-                        .clone()
-                        .try_into()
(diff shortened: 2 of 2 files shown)


poseidon2-circuit-air/src/public_types.rs
```diff
@@ -31,6 +32,7 @@ impl Poseidon2Params for BabyBearD4Width24 {
     const WIDTH: usize = 24;
     const RATE_EXT: usize = 4;
     const CAPACITY_EXT: usize = 2;
+    const DIGEST_EXT: usize = 4;
     const SBOX_DEGREE: u64 = 7;
     const SBOX_REGISTERS: usize = 1;
     const HALF_FULL_ROUNDS: usize = 4;
```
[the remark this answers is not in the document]
Nashtare: Yeah I fixed it while adding the test

poseidon2-circuit-air/src/public_types.rs
```diff
@@ -33,6 +34,7 @@ impl Poseidon2Params for BabyBearD4Width24 {
     const WIDTH: usize = 24;
     const RATE_EXT: usize = 4;
     const CAPACITY_EXT: usize = 2;
+    const DIGEST_EXT: usize = 2;
```
hratoanina: Is DIGEST_EXT still 2 in a 24-width permutation?

poseidon2-circuit-air/src/air.rs
```diff
@@ -201,40 +206,71 @@ impl<
...
-                                state[D..2 * D].copy_from_slice(&prev_out[D..2 * D]);
+                        let cur_bit = *mmcs_bit;
+                        // For trace generation, we chain unconditionally here.
+                        // The AIR will enforce that chaining only applies when in_ctl = 0.
+                        if cur_bit {
+                            // Case B: mmcs_bit = 1 (right = previous hash)
+                            // If RATE_EXT >= 2*DIGEST_EXT, chain from second half of rate section
+                            // Otherwise, fall back to capacity limbs
+                            if RATE_EXT >= 2 * DIGEST_EXT {
+                                // Chain from second half of rate: in_{r+1}[0..DIGEST_EXT-1] = out_r[DIGEST_EXT..2*DIGEST_EXT-1]
```
hratoanina: I think the source is still `out_r[0..DIGEST_EXT]`, but it's copied into `in_{r+1}[DIGEST_EXT..2*DIGEST_EXT-1]`. I'm surprised it's not caught by any test.

I'm not sure how pertinent it is to correct it though, since the state computation logic will be removed from here in https://github.com/Plonky3/Plonky3-recursion/pull/222.

PR #204, opened 2025-12-03
Add Poseidon Perm Op

sai-deng:
- Replace hash absorb/squeeze with a single PoseidonPerm op using extension witnesses (D=4), with per-row flags carried as params and mmcs_bit as a witness value.
  - Update Poseidon trace builder/AIR to downcast the executor for flags, honor in_ctl gating, enforce mmcs_bit booleanity, and align chaining/accumulator semantics.
  - Adjust prover glue to accept extension traces and gate D=4 only on the d4 path; keep registration opt-in.
  - Add Poseidon perm chain and Merkle examples (Merkle now 3 rows, public root/index), drop hash_chain, and clean up stale TODOs.
  - Tests/examples updated; full test suite passes.

### DIFF
--- a/circuit/src/ops/poseidon_perm.rs
+++ b/circuit/src/ops/poseidon_perm.rs
@@ -1,4 +1,17 @@
 //! Poseidon permutation non-primitive operation (one Poseidon call per row).
+//!
+//! This operation is designed to support both standard hashing and specific logic required for
+//! Merkle path verification within a circuit. Its features include:
+//!
+//! - **Hashing**: Performs a standard Poseidon permutation.
+//! - **Chaining**: Can start a new hash computation or continue from the output of the previous row
+//!   (controlled by `new_start`).
+//! - **Merkle Path Verification**: When `merkle_path` is enabled, it supports logic for verifying
+//!   a path up a Merkle tree. This involves conditionally arranging inputs (sibling vs. computed hash)
+//!   based on a direction bit (`mmcs_bit`).
+//! - **Index Accumulation**: Supports accumulating path indices (`mmcs_index_sum`) to verify the
+//!   leaf's position in the tree.
+//!
 //! Only supports extension degree D=4 for now.
 
 use alloc::boxed::Box;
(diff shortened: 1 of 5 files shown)


circuit/src/ops/poseidon_perm.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// User-facing arguments for adding a Poseidon perm row.
+pub struct PoseidonPermCall {
+    pub new_start: bool,
+    pub merkle_path: bool,
+    /// Optional mmcs direction bit input (base field, boolean). If None, defaults to 0/private.
+    pub mmcs_bit: Option<ExprId>,
+    /// Optional CTL exposure for each input limb (one extension element).
+    /// If `None`, the limb is considered private/unexposed (in_ctl = 0).
+    pub inputs: [Option<ExprId>; 4],
```
4l0n50: We won't be supporting other configs for now?

circuit/src/ops/poseidon_perm.rs
```diff
@@ -0,0 +1,156 @@
+//! Poseidon permutation non-primitive operation (one Poseidon call per row).
```
LindaGuiga: Since this Op deals with both hashing and Merkle path verification, it might be worth explaining everything this Op can do

circuit/src/tables/poseidon2.rs
```diff
@@ -112,27 +114,31 @@ impl<F: Clone + Send + Sync + 'static> NonPrimitiveTrace<F> for Poseidon2Trace<F
...
+    fn boxed_clone(&self) -> Box<dyn NonPrimitiveTrace<CF>> {
         Box::new(self.clone())
     }
 }
 
 /// Builder for generating Poseidon2 traces.
-pub struct Poseidon2TraceBuilder<'a, F, Config: Poseidon2Params> {
-    circuit: &'a Circuit<F>,
-    witness: &'a [Option<F>],
+pub struct Poseidon2TraceBuilder<'a, CF, Config: Poseidon2Params> {
```
LindaGuiga: Why this name change in this PR?

LindaGuiga (approved):
LGTM, thank you!

poseidon2-circuit-air/src/air.rs
```diff
@@ -446,7 +461,7 @@ fn eval<
             .assert_zero(next_in[d].clone() - local_out[d].clone());
 
         // Right case: in_{r+1}[0] = out_r[2]
-        let gate_right_0 = next.merkle_chain_sel[0].clone() * prev_bit.clone();
+        let gate_right_0 = next.merkle_chain_sel[0].clone() * next_bit.clone();
```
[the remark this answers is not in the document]
sai-deng: `next.mmcs_bit`: Specifies the structure of this Merkle node (row i+1). It determines whether the previous hash (local output) should be placed in the "Left" or "Right" input position of the new hash.

If we used local.mmcs_bit, we would be using the child's direction bit to configure the parent's input layout, which is incorrect.

PR #205, opened 2025-12-03
circuit: add unit tests and doc for `generate_preprocessed_columns`

tcoratger:


circuit/src/circuit.rs
```diff
@@ -94,27 +95,93 @@ impl<F: Field> Circuit<F> {
...
 
-    /// Generates the preprocessed values for all ops except non-primitive ops.
+    /// Generates preprocessed columns for all primitive operation types.
     ///
-    /// The preprocessed values for `Witness` are deduced from the other ops:
-    /// they correspond to 0..`n` where `n` is the largest witness index used in the circuit.
-    pub fn generate_preprocessed_columns(&self) -> Result<Vec<Vec<F>>, crate::CircuitError> {
-        let n = PrimitiveOpType::COUNT; // Exclude non-primitive ops
-        let mut preprocessed = vec![vec![]; n];
+    /// # Overview
```
LindaGuiga: Overall, I feel like the doc for this method is too long: we would have to have similar docs for methods in general to keep things unified, and it could quickly become difficult to maintain. Maybe we could put the preprocessing explanation as a header in a file (not necessarily `circuit`) or directly in the mdbook, since it corresponds to a global explanation of how the recursive circuit works/manages preprocessing?

LindaGuiga (commented):
Thank you for this work!

Overall, I feel like this is adding too much code/documentation (+702 lines) for just one method. Since we would have to unify it with the rest of the repo, I am afraid this could quickly become hard to manage.

As mentioned below, regarding the doc: maybe we could add it to a header somewhere to explain how preprocessing works in general in our recursive circuit (or directly in the mdbook)?

Regarding the tests, I feel like they are a bit redundant sometimes: for example, I don't think we need both `test_single_const_populates_const_column_with_output_index` and `test_multiple_consts_append_indices_in_order` (just the latter should suffice). Maybe we could have just 2 or 3 good unit tests to avoid bloating the repo's code (and therefore also making maintaining the code harder/ future PRs bigger)?

Let me know what you think!

PR #210, opened 2025-12-08
[feat] Implement recursive Mmcs verify batch

4l0n50:
This PR introduce a recursive variant of `mmcs::verify_batch` and a circuit-side `hash_slice` helper that mirrors `CryptographicHasher::hash_iter`, both implemented via chained Poseidon2 permutations.

Bugs fixed along the way:
- Poseidon2 merkle-path input resolution to match recursive MMCS semantics:

### DIFF
--- a/recursion/src/pcs/mmcs.rs
+++ b/recursion/src/pcs/mmcs.rs
@@ -1,5 +1,5 @@
-use alloc::vec::Vec;
 use alloc::format;
+use alloc::vec::Vec;
 
 use p3_circuit::op::{NonPrimitiveOpType, Poseidon2Config};
 use p3_circuit::ops::hash::add_hash_slice;
--- a/circuit/src/builder/errors.rs
+++ b/circuit/src/builder/errors.rs
@@ -75,5 +75,8 @@ pub enum CircuitBuilderError {
 
     /// Failed to format openings for MMCS preprocessing; preserves some context.
     #[error("Failed to format openings for operation {op:?}: {details}")]
-    FormatOpeningsFailed { op: NonPrimitiveOpType, details: String },
+    FormatOpeningsFailed {
+        op: NonPrimitiveOpType,
+        details: String,
+    },
 }


circuit/src/ops/poseidon2_perm.rs
```diff
@@ -604,8 +608,8 @@ impl<F: Field + Send + Sync + 'static> NonPrimitiveExecutor<F> for Poseidon2Perm
 
 impl Poseidon2PermExecutor {
     /// Resolve input limb value using a layered priority system:
-    /// 1. Layer 1: Private inputs (lowest priority) - sibling placed based on mmcs_bit
-    /// 2. Layer 2: Chaining from previous permutation (if new_start=false)
+    /// 1. Layer 1: Chaining from previous permutation (lowest priority) or zeros if new_start
+    /// 2. Layer 2: Private inputs - sibling placed based on mmcs_bit
     /// 3. Layer 3: CTL (witness) values (highest priority, overwrites previous layers)
```
Nashtare: The reordering on priorities from `private → chaining → CTL` to `chaining → private → CTL` is a bit unclear to me, why is that?

circuit/src/ops/hash.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::op::Poseidon2Config;
+use crate::ops::Poseidon2PermCall;
+use crate::ops::poseidon2_perm::Poseidon2PermOps;
+use crate::{CircuitBuilder, CircuitBuilderError, ExprId, NonPrimitiveOpId};
+
+pub fn add_hash_slice<F: Field>(
+    builder: &mut CircuitBuilder<F>,
+    poseidon2_config: &Poseidon2Config,
+    inputs: &[ExprId],
+    reset: bool,
```
Nashtare: So the `reset` flag is only used for the first chunk right? I guess this is fine as restricted to MMCS but if we wanted to say, absorb into an existing sponge state with this method, we may need to tweak it

recursion/src/pcs/mmcs.rs
```diff
@@ -0,0 +1,492 @@
...
+        opened_values,
+        dimensions,
+        index_bits.len(),
+        permutation_config,
+    )
+    .map_err(
+        |_| CircuitBuilderError::InvalidNonPrimitiveOpConfiguration {
+            // TODO: I this the error we want?
+            op: NonPrimitiveOpType::Poseidon2Perm(permutation_config),
+        },
```
Nashtare: why not just extracting the underlying error?

circuit/src/ops/mmcs.rs
```diff
@@ -0,0 +1,126 @@
...
+        .peekable();
+
+    // Matrix heights that round up to the same power of two must be equal
+    if !heights_tallest_first
+        .clone()
+        .map(|(_, dims)| dims.height)
+        .tuple_windows()
+        .all(|(curr, next)| curr == next || curr.next_power_of_two() != next.next_power_of_two())
+    {
+        panic!("Heights that round up to the same power of two must be equal"); //TODO: Add errors
```
Nashtare: probably best not to panic indeed

circuit/src/ops/hash.rs
Nashtare: may also deserve a dedicated test for `reset = false`?

4l0n50: This is supposed to behave as `CryptographicHasher::hash_iter` plus the optional reset. I think what you are saying is equivalent to `add_hash_slice(builder, config, inputs_that_will_lead_to_existing_sponge_state, false); add_hash_slice(builder, config, new_inputs_to_absorb_into_existing_sponge_state, false)`. So, if I'm understanding what you said correctly, we shouldn't need to tweak.

circuit/src/ops/poseidon2_perm.rs
4l0n50: Because you want to overwrite the previous state (chaining) with the private input (the sibling), not the other way

PR #211, opened 2025-12-09
circuit builder: move bit decomposition

tcoratger:


circuit/src/utils.rs
```diff
@@ -107,107 +104,6 @@ pub fn symbolic_to_circuit<F: Field>(
...
-) -> Result<Vec<ExprId>, CircuitError> {
-    builder.push_scope("decompose_to_bits");
-
-    // Create bit witness variables
-    let binary_decomposition_hint = BinaryDecompositionHint::new(x, n_bits)?;
-    let bits = builder.alloc_witness_hints(binary_decomposition_hint, "decompose_to_bits");
-
-    for &bit in bits.iter() {
-        builder.assert_bool(bit);
-    }
```
tcoratger: When migrating to the circuit builder, I removed this loop because this is also the first step done in the `reconstruct_index_from_bits` method and so that is redundant I think.

Nashtare (approved):

PR #213, opened 2025-12-12
Implement `AirLookupHandler` for primitive and non-primitive operations

LindaGuiga:
This PR aims at implementing `AirLookupHandler` for both primitive and non-primitive operations. Lookups are, however, not yet integrated to the circuit verifier. 

Note that in `PoseidonCircuitAir`, we could group the lookups 2 by 2 (similarly to `AddAir` for example). However, since we can't test the verification yet, I thought it would be better for later debugging to leave this simpler version for now.

### DIFF
--- a/recursion/tests/air_symbolic_to_circuit.rs
+++ b/recursion/tests/air_symbolic_to_circuit.rs
@@ -144,13 +144,23 @@ fn primitive_airs_symbolic_to_circuit() -> Result<(), CircuitError> {
     run_recursive(&mul_air, mul_air.preprocessed_width(), 0, &mut rng)?;
 
     let const_air = ConstAir::<F, 1>::new_with_preprocessed(1, vec![F::from_u64(3)]);
-    run_recursive(&const_air, 1, 0, &mut rng)?;
+    run_recursive(
+        &const_air,
+        ConstAir::<F, 1>::preprocessed_width(),
+        0,
+        &mut rng,
+    )?;
 
     let public_air = PublicAir::<F, 1>::new_with_preprocessed(1, vec![F::from_u64(4)]);
-    run_recursive(&public_air, 1, 1, &mut rng)?;
+    run_recursive(
+        &public_air,
+        PublicAir::<F, 1>::preprocessed_width(),
+        1,
+        &mut rng,
+    )?;
 
     let witness_air = WitnessAir::<F, 1>::new(1, 1);
-    run_recursive(&witness_air, 1, 0, &mut rng)?;
+    run_recursive(&witness_air, witness_air.preprocessed_width(), 0, &mut rng)?;
 
     Ok(())
 }


poseidon2-circuit-air/src/air.rs
```diff
@@ -583,6 +586,152 @@ impl<
...
+            ));
+        }
+
+        // If next.new_start = 1 and local.merkle_path = 1, then mmcs_index_sum is exposed via CTL.
+        let multiplicity = next.new_start * local.merkle_path;
+        let mmcs_index_sum_lookup = [
+            SymbolicExpression::from(local.mmcs_index_sum_idx),
+            SymbolicExpression::from(local.mmcs_index_sum),
+            // Since `mmcs_index_sum` is a single field element (not an extension), its other limbs are zero.
+            SymbolicExpression::Constant(AB::F::ZERO),
```
sai-deng: This is working for now (since we always use D=4). However, it should be easy to implement it based on D.

circuit-prover/src/air/utils.rs
```diff
@@ -0,0 +1,118 @@
...
+        self.get_lookups()
+    }
+    fn eval_lookups_dyn(
+        &self,
+        builder: &mut AB,
+        contexts: &[Lookup<AB::F>],
+        lookup_data: &[LookupData<AB::EF>],
+        gadget: &dyn LookupGadgetDyn<AB>,
+    ) {
+        Air::<AB>::eval(self, builder);
```
sai-deng: I am wondering if eval() will be called twice?

circuit/src/circuit.rs
```diff
@@ -103,34 +102,48 @@ impl<F: Field> Circuit<F> {
...
     /// |-------|-----------|-------------------------------------------|----------------|
-    /// | 0     | Witness   | `[idx_0, idx_1, ..., idx_max]`            | 1              |
+    /// | 0     | Witness   | `[idx_0, mul_0, idx_1, mul_1, ...]`       | 2              |
     /// | 1     | Const     | `[out_0, out_1, ...]`                     | 1              |
     /// | 2     | Public    | `[out_0, out_1, ...]`                     | 1              |
     /// | 3     | Add       | `[a_0, b_0, out_0, a_1, b_1, out_1, ...]` | 3              |
     /// | 4     | Mul       | `[a_0, b_0, out_0, a_1, b_1, out_1, ...]` | 3              |
+    ///
+    /// Note that `mul_i` in the Witness table preprocessed column indicates how many times
+    /// each witness index appears in the circuit.
```
Nashtare: If `mul_N` are preprocessed columns, shouldn't we change the `get_prep_width` function? It still returns only 1 per lane (although granted we don't use it outside of tests, but I'd suggest we keep it up to date or fully remove it to make sure there's no confusion).

circuit/src/op.rs
```diff
@@ -373,6 +373,14 @@ pub trait NonPrimitiveExecutor<F: Field>: Debug {
         ctx: &mut ExecutionContext<'_, F>,
     ) -> Result<(), CircuitError>;
 
+    /// Update the preprocessed columns for this operation, as well as for `Witness` table.
```
Nashtare: The function doc is a bit unclear

circuit/src/ops/poseidon_perm.rs
```diff
@@ -157,6 +157,38 @@ impl<F: Field> NonPrimitiveExecutor<F> for PoseidonPermExecutor {
...
+            for witness_id in out {
+                let idx = witness_id.0 as usize;
+                if idx >= preprocessed_tables[witness_table_idx].len() {
+                    preprocessed_tables[witness_table_idx].resize(idx + 1, F::from_u32(0));
+                }
+                preprocessed_tables[witness_table_idx][idx] += F::ONE;
+            }
+        }
+
+        // TODO: Update preprocessing columns for the Permutation table as well.
```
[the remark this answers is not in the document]
Nashtare: Ah I just saw the `"Lookups in dynamic tables not supported yet"` comment, I assume it's related? Would be nice to have a TODO next to it for tracking if its out of scope of this PR (its title is a bit misleading in this regard).

circuit/src/circuit.rs
```diff
@@ -103,34 +102,48 @@ impl<F: Field> Circuit<F> {
...
     pub fn generate_preprocessed_columns(&self) -> Result<Vec<Vec<F>>, CircuitError> {
```
Nashtare: Can be left for later I guess, but I think we could do some refactoring here by centralizing the multiplicity logic to prevent any miscount and simplify the read/write accounting.

circuit-prover/src/air/add_air.rs
```diff
@@ -395,6 +406,65 @@ where
...
+        );
+
+        let symbolic_main = symbolic_air_builder.main();
+        let symbolic_main_local = symbolic_main.row_slice(0).unwrap();
+
+        let preprocessed = symbolic_air_builder.preprocessed();
+        let preprocessed_local = preprocessed.row_slice(0).unwrap();
+
+        // We use the same multiplicity for all lookups.
+        let multiplicity = SymbolicExpression::Constant(AB::F::ONE);
```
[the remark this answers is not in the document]
Nashtare: (would be worth having a small unit test ensuring padded lanes do contain multiplicity 0, including also half-padded ones).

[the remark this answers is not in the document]
sai-deng: We always have (id, value) = (0, 0) in the first row of the witness table. So if we pad with 0 + 0 = 0 or 0 * 0 = 0, it should work regardless of whether it’s included in the CTL.

circuit/src/circuit.rs
LindaGuiga: I actually removed `get_prep_width`, as it wasn't used anywhere, I believe (I didn't see a test using it either...)

circuit/src/circuit.rs
LindaGuiga: I added a TODO

PR #216, opened 2025-12-16
refactor: Unified Op Execution

sai-deng:


circuit/src/builder/compiler/expression_lowerer.rs
```diff
@@ -123,6 +257,43 @@ where
...
     > {
+        // Precompute mapping from op_id -> output expression nodes, to allow emitting an op once
+        // while still producing witness ids for all of its outputs.
+        let mut op_id_to_output_exprs: HashMap<u32, Vec<(u32, usize, ExprId)>> = HashMap::new();
+        for (expr_idx, expr) in self.graph.nodes().iter().enumerate() {
+            if let Expr::NonPrimitiveOutput { op_id, output_idx } = expr {
+                op_id_to_output_exprs.entry(op_id.0).or_default().push((
+                    *output_idx,
+                    expr_idx,
+                    ExprId(expr_idx as u32),
```
4l0n50: why do you add twice `expr_idx`, and one wrapped as `ExprId`?

circuit/src/expr.rs
```diff
@@ -22,6 +22,19 @@ pub enum Expr<F> {
     Mul { lhs: ExprId, rhs: ExprId },
     /// Division of two expressions
     Div { lhs: ExprId, rhs: ExprId },
+    /// Anchor node for a non-primitive operation in the expression DAG.
+    ///
+    /// This node has no witness value itself, but it fixes the relative execution order
+    /// of non-primitive ops w.r.t. other expressions during lowering.
+    NonPrimitiveCall { op_id: NonPrimitiveOpId },
```
4l0n50: Why not requiring  `inputs: Vec<ExprId>` here? that would be more in line with the DAG structure of other non-degree 0 nodes

sai-deng: When I add it, it serves as a marker in the DAG to enforce the execution order relative to other expressions. The inputs are already stored in the NonPrimitiveOperationData struct, and including them in Expr would duplicate that data. However, I agree that having the inputs here could be a better design if we add DAG optimizations in the future.

circuit/src/builder/compiler/expression_lowerer.rs
```diff
@@ -279,14 +449,60 @@ where
...
+                            &mut ops,
+                        )?;
+                    }
+                }
+                Expr::NonPrimitiveOutput {
+                    op_id,
+                    output_idx: _,
+                } => {
+                    if emitted_non_primitive_ops.insert(op_id.0) {
+                        let data = self
```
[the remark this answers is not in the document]
sai-deng: Even if the NonPrimitiveCall node exists and was processed earlier (emitting the Op and populating the witness map), the iterator will still eventually reach the NonPrimitiveOutput node. We need this branch to be valid (not an error) so the loop can gracefully pass over these nodes.

sai-deng:
Thanks for the quick review @4l0n50 ! I’ve merged the PR, but please feel free to leave follow-up comments—I can address them in the next PR.

PR #218, opened 2025-12-16
Enable preprocessing for non primitive operations

LindaGuiga:
This PR aims at enabling preprocessing for non-primitive operations, which includes implementing `preprocessed_trace` for `Poseidon2CircuitAir`.

### DIFF
--- a/circuit/src/op.rs
+++ b/circuit/src/op.rs
@@ -360,23 +360,15 @@ pub trait NonPrimitiveExecutor<F: Field>: Debug {
         ctx: &mut ExecutionContext<'_, F>,
     ) -> Result<(), CircuitError>;
 
-    /// Update the preprocessed values related to this operation. This consists of:
-    /// - the preprocessed values for the associated table
-    /// - the multiplicity for the `Witness` table.
-    fn preprocessing(
-        &self,
-        inputs: &[Vec<WitnessId>],
-        outputs: &[Vec<WitnessId>],
-        preprocessed_tables: &mut Vec<Vec<F>>,
-    );
-
     /// Get operation type identifier (for config lookup, error reporting)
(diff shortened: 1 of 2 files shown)


circuit/src/circuit.rs
```diff
@@ -163,12 +163,25 @@ impl<F: Field> Circuit<F> {
...
-                Op::NonPrimitiveOpWithExecutor { .. } => panic!(
-                    "preprocessed values are not yet implemented for non primitive operations."
+                Op::NonPrimitiveOpWithExecutor { executor, .. } => panic!(
+                    "Unexpected non-primitive op {:?} in primitive_ops",
+                    executor.op_type()
                 ),
             }
         }
 
+        for nop in &self.non_primitive_ops {
```
4l0n50: op? nop sounds like no operation

circuit/src/op.rs
```diff
@@ -379,6 +379,14 @@ pub trait NonPrimitiveExecutor<F: Field>: Debug {
     /// Allow downcasting to concrete executor types
     fn as_any(&self) -> &dyn core::any::Any;
 
+    /// Update the preprocessed columns with the values added by this operation.
+    fn preprocessing(
```
4l0n50: `execute_preprocessing()`?

circuit/src/op.rs
```diff
@@ -379,6 +379,14 @@ pub trait NonPrimitiveExecutor<F: Field>: Debug {
...
+        &self,
+        inputs: &[Vec<WitnessId>],
+        outputs: &[Vec<WitnessId>],
+        preprocessed_tables: &mut Vec<Vec<F>>,
```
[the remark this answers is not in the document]
LindaGuiga: The problem is that we also have to update the `Witness` multiplicities (see PR #213 ) for CTLs.

PR #221, opened 2025-12-22
fix: make dynamic dispatch delegate to inner AIR logic

### LINKED ISSUE #220: bug: dynamic dispatch skips actual table constraint evaluation
Mentioned by @hratoanina 

Running any poseidon example with invalid AIR constraints for poseidon2 table (or even a `panic!()` inside) will pass fine. Seems to be caused by dynamic dispatch logic

Nashtare:
# Description

We were not delegating to the actual inner AIRs when performing dynamic dispatch of non-primitive operations, which would make anything pass without any actual checks.

# Changes

I've implemented `eval` for `Poseidon2AirWrapper`. It was messy because of type inferences and Rust limitations so unlike primitive tables, the `eval` implementation had to be duplicated: one for `SymbolicAirBuilder`, one for `ProverConstraintFolder`, etc...
I tried wrapping most of redundancies together to reduce duplication.

closes #220

### DIFF
--- a/circuit-prover/src/batch_stark_prover.rs
+++ b/circuit-prover/src/batch_stark_prover.rs
@@ -476,6 +476,17 @@ enum Poseidon2AirWrapperInner {
     KoalaBearD4Width24(Box<Poseidon2CircuitAirKoalaBearD4Width24>),
 }
 
+impl Poseidon2AirWrapperInner {
+    fn width(&self) -> usize {
+        match self {
+            Self::BabyBearD4Width16(air) => air.width(),
+            Self::BabyBearD4Width24(air) => air.width(),
+            Self::KoalaBearD4Width16(air) => air.width(),
+            Self::KoalaBearD4Width24(air) => air.width(),
+        }
+    }
+}
+
 struct Poseidon2AirWrapper<SC: StarkGenericConfig> {
     inner: Poseidon2AirWrapperInner,
     width: usize,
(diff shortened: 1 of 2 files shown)


circuit-prover/src/batch_stark_prover.rs
```diff
@@ -871,6 +1598,7 @@ where
                 use p3_poseidon2_circuit_air::Poseidon2CircuitAirBabyBearD4Width16;
                 let air = Poseidon2CircuitAirBabyBearD4Width16::new(constants.clone());
```
sai-deng: This creates an AIR for width(), then creates a different AIR via to_air_wrapper(). Could we use the same instance to avoid potential state inconsistency?

circuit-prover/src/batch_stark_prover.rs
```diff
@@ -476,37 +513,722 @@ where
...
+impl<SC> AirLookupHandler<SymbolicAirBuilder<Val<SC>>> for Poseidon2AirWrapper<SC>
+where
+    SC: StarkGenericConfig + Send + Sync,
+    Val<SC>: StarkField + PrimeField,
+{
+    fn add_lookup_columns(&mut self) -> Vec<usize> {
+        match &mut self.inner {
+            Poseidon2AirWrapperInner::BabyBearD4Width16(air) => unsafe {
+                let air_bb: &mut Poseidon2CircuitAirBabyBearD4Width16 =
+                    core::mem::transmute(air.as_mut());
```
sai-deng: Is this transmute needed? air.as_mut() already returns the correct type.

circuit-prover/src/batch_stark_prover.rs
```diff
@@ -476,37 +513,722 @@ where
...
+
+    #[allow(clippy::missing_transmute_annotations)] // this gets overly verbose otherwise
+    fn get_lookups(&mut self) -> Vec<p3_lookup::lookup_traits::Lookup<Val<SC>>> {
+        match &mut self.inner {
+            Poseidon2AirWrapperInner::BabyBearD4Width16(air) => unsafe {
+                let air_bb: &mut Poseidon2CircuitAirBabyBearD4Width16 =
+                    core::mem::transmute(air.as_mut());
+                let lookups_bb = <Poseidon2CircuitAirBabyBearD4Width16 as AirLookupHandler<
+                    SymbolicAirBuilder<BabyBear>,
+                >>::get_lookups(air_bb);
```
sai-deng: Should we add a runtime guard ensure Val<SC> matches the field type?

poseidon2-circuit-air/src/air.rs
```diff
@@ -434,7 +437,7 @@ pub fn extract_preprocessed_from_operations<F: Field, OF: Field>(
     preprocessed
 }
 
-fn eval<
+pub fn eval<
```
sai-deng: Is pub fn eval intentionally public API? If only eval_unchecked needs external visibility, could this be pub(crate)?

sai-deng (approved):

PR #222, opened 2025-12-23
refactor: Unify Poseidon execution and trace generation

sai-deng:
Consolidates Poseidon2 permutation code into circuit/src/ops/poseidon2_perm.rs and makes ExecutionContext store per‑op state via a type‑erased OpStateMap.

  Key changes:
  - Replace last_poseidon with op_states: BTreeMap<NonPrimitiveOpType, Box<dyn OpExecutionState>>.
  - Record Poseidon2CircuitRow during execution and generate traces from recorded rows, removing chaining logic from trace generation.
  - Move Poseidon2PermConfig, Poseidon2PermPrivateData, Poseidon2Trace, and related types into ops/poseidon2_perm.rs.
  - Delete circuit/src/tables/poseidon2.rs.

### DIFF
--- a/circuit-prover/src/common.rs
+++ b/circuit-prover/src/common.rs
@@ -1,3 +1,4 @@
+use alloc::collections::btree_map::BTreeMap;
 use alloc::vec::Vec;
 
 use p3_circuit::op::{NonPrimitiveOpType, Poseidon2Config, PrimitiveOpType};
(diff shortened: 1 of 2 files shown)


poseidon2-circuit-air/src/air.rs
```diff
@@ -136,7 +136,7 @@ impl<
 
     pub fn generate_trace_rows<P: CryptographicPermutation<[F; WIDTH]>>(
```
hratoanina: Shouldn't the trace generation logic here be removed?

circuit/src/ops/poseidon2_perm.rs
```diff
@@ -16,20 +23,101 @@
...
+
+/// Configuration for Poseidon2 permutation operations.
+///
+/// Contains an execution closure that computes the Poseidon2 permutation.
+/// The closure takes 4 extension field limbs and returns 4 output limbs.
+pub struct Poseidon2PermConfig<F> {
+    /// Execution closure: converts [F;4] extension limbs to [Base;16],
+    /// runs the permutation, and converts back to [F;4].
+    pub exec: Poseidon2PermExec<F>,
+}
```
Nashtare: Could we at least start adding some `const WIDTH: usize, const DIGEST: usize` to its associated types (even if this PR will just use 16 / 4) to reduce overhead later on with generic poseidon?

circuit/src/ops/poseidon2_perm.rs
```diff
@@ -16,20 +23,101 @@
...
+// ============================================================================
+// Private Data
+// ============================================================================
+
+/// Private data for Poseidon2 permutation.
+/// Only used for Merkle mode operations, contains exactly 2 extension field limbs (the sibling).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Poseidon2PermPrivateData<F> {
+    pub sibling: [F; 2],
+}
```
Nashtare: Same here for new types/structs

circuit/src/op.rs
```diff
@@ -322,9 +294,30 @@ impl<F> Hash for NonPrimitiveOpConfig<F> {
...
+    fn as_any(&self) -> &dyn Any;
+    /// Downcast to concrete type for mutation.
+    fn as_any_mut(&mut self) -> &mut dyn Any;
 }
 
+/// Type-erased storage for operation execution states.
+///
+/// This allows each operation type to maintain its own state without
+/// coupling `ExecutionContext` to specific operation implementations.
+pub type OpStateMap = BTreeMap<NonPrimitiveOpType, Box<dyn OpExecutionState>>;
```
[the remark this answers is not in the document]
sai-deng: Good point! Addressed: NonPrimitiveOpType now carries the Poseidon2Config, so the map key is unique per config. Poseidon2Perm(BabyBearD4Width16) and Poseidon2Perm(BabyBearD4Width24) map to different entries, so both configs can coexist.

circuit/src/tables/runner.rs
```diff
@@ -162,11 +167,16 @@ impl<F: CircuitField> CircuitRunner<F> {
 
         let mut non_primitive_traces: HashMap<&'static str, Box<dyn NonPrimitiveTrace<F>>> =
             HashMap::new();
-        for generator in self.circuit.non_primitive_trace_generators.values() {
+        // Iterate over generators in deterministic order (sorted by key)
```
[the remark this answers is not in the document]
sai-deng: We sort to keep trace generation deterministic across hash map iteration order. That makes runs/replays stable and avoids non‑reproducible traces/proofs when map iteration order changes.

circuit-prover/src/common.rs
```diff
@@ -67,120 +67,102 @@ where
...
-                            poseidon2_wrapper_air,
-                            log2_ceil_usize(prep.len().div_ceil(width)),
-                        );
+            Ok(())
+        })?;
 
-                        Ok(())
-                    }
-                    _ => panic!("Unknown primitive operation at index {}", primitive_idx),
+    if let Some(configs) = non_primitive_configs {
```
LindaGuiga: Nit: I'm just wondering: wouldn't it be better to loop on preprocessed.non_primitive directly? I'm just thinking that we might want to give the configuration in case there are some such operations, but there might be none. For example, we provide a Poseidon2 configuration just in case there are some hashes in the circuit, but if we have an empty proof or something like that, there might end up being no Poseidon2 operation, and we could get rid of the table. Of course, for now we only have Poseidon2 and so we should always have hashes in our recursive circuit, but I'm thinking that we (or other users) might add tables that can end up being empty

sai-deng: Done. We now iterate preprocessed.non_primitive and only build tables for ops that actually appear.

PR #224, opened 2025-12-24
chore: fix naming for Poseidon2 code

Nashtare:
No functionality change, just renaming `poseidon` -> `poseidon2`, to match the actual specs and prevent any confusion (Poseidon and Poseidon2 being two different families).


LindaGuiga (approved):
LGTM, thanks!

PR #245, opened 2026-02-01
refactor: RFC-0001

### LINKED ISSUE #234: [Task] Refactor witness multiplicities handling
### DRI (owner)

@LindaGuiga

### Flags

- [ ] perf
- [x] needs-rfc

### RFC link (required if needs-rfc is checked)

https://github.com/Plonky3/Plonky3-recursion/blob/main/docs/rfcs/0001-rfc-handle-multiplicities.md

### Problem / Motivation

With lookups and preprocessing, we introduced preprocessed data and witness multiplicities. 
Currently, witness multiplicities are updated manually (similarly to other processed data), which makes things a bit error-prone: we can easily forget to update the multiplicities when adding a column in another table. Thus, it would be better to handle multiplicities automatically.

Another issue that #225 introduces is that we need to pass the multiplicities to the prover, which is not user-friendly.

Finally, preprocessed data and multiplicities are also padded outside of the `generate_preprocessed_data` method, when `Add` or `Mul` operations are empty. Once again, this is error-prone.

### Scope

In:
- Update `generate_preprocessed_data` and other preprocessed updates if necessary, to make things less error-prone
- We might need to change `CommonData` (or have `ProverDataOnly`/`VerifierDataOnly`)

Out:
- We might improve how all preprocessed data in general is generated

### Proposed approach

We could introduce a `read_from_witness` method that automatically updates preprocessed data accordingly (adds the witness index to the table's preprocessed data and updates witness multiplicities).

We could also update common.rs and divide it into `ProverOnlyData`/`VerifierOnllyData` so that we can store `witness_multiplicities` in `ProverOnlyData` and pass that to the prover, to make things more user-friendly.

### Exit criteria

- Refactor complete 
- Tests passing

### Risk / Rollout

_No response_

### Dependencies / Blockers

_No response_

### Notes / Links

_No response_

Nashtare:
## Summary

Refactors multiplicities according to RFC-0001.

Fixes #234.

### DIFF
--- a/circuit-prover/src/batch_stark_prover.rs
+++ b/circuit-prover/src/batch_stark_prover.rs
@@ -3872,7 +3872,7 @@ mod tests {
         let proof = prover
             .prove_all_tables(&traces, &circuit_prover_data)
             .unwrap();
-        prover.verify_all_tables(&proof, &common).unwrap();
+        prover.verify_all_tables(&proof, common).unwrap();
     }
 
     #[test]
@@ -3919,6 +3919,6 @@ mod tests {
         let proof = prover
             .prove_all_tables(&traces, &circuit_prover_data)
             .unwrap();
-        prover.verify_all_tables(&proof, &common).unwrap();
+        prover.verify_all_tables(&proof, common).unwrap();
     }
 }
--- a/recursion/tests/test_lookups.rs
+++ b/recursion/tests/test_lookups.rs
@@ -875,7 +875,7 @@ fn test_poseidon2_ctl_lookups() {
         .unwrap();
 
     prover
-        .verify_all_tables(&proof, &common)
+        .verify_all_tables(&proof, common)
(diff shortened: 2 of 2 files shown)


circuit/src/circuit.rs
```diff
@@ -46,6 +46,136 @@ pub struct PreprocessedColumns<F> {
...
+        &mut self,
+        op_type: PrimitiveOpType,
+        wids: &[WitnessId],
+    ) -> Result<(), CircuitError> {
+        if self.primitive.len() != PrimitiveOpType::COUNT {
+            return Err(CircuitError::InvalidPreprocessing {
+                reason: "primitive vector length does not match PrimitiveOpType::COUNT",
+            });
+        }
+
```
LindaGuiga: Should we also check here that the op type is not `Witness` (so we don't update the witness table twice)?

PR #251, opened 2026-02-05
[RFC] Witness removal

### LINKED ISSUE #103: Optimize Witness Table
This issue may involve:

- Eliminating the witness table via additional in-table and cross-table lookups. This requires wiring all LHS/RHS/RES columns for primitive op tables (ADD, MUL, …) internally and connecting them to the relevant non-primitive tables.

- Making the witness table multi-lane, similar to #81.

LindaGuiga:
## Summary
This RFC proposes a way to remove the Witness table, while allowing the challenger to handle base field as well as extension field elements. 

Fixes #103 

## Why
The Witness table is currently the largest, and the information it contains is redundant with the indices stored in all other AIRs. Thus, getting rid of it while keeping the soundness guarantees should improve the performance by a lot.

Note that this RFC also proposes a way to handle both extension field elements and extension field elements for the challenger. This leads to both extension field and base field CTLs. I guess we could simplify this by always having base field CTLs, but I think that would lead to more CTLs.

## Changes
- [ ] This RFC proposes a way to remove the Witness table. This also leads to multiple other changes, including preprocessing, indexing of the targets, and CTLs.

## Acceptance criteria
- [ ] The proposal is deemed functional and likely to bring performance gains.

## Tests
- [ ] Unit tests added/updated
- [x] Existing tests pass
- [ ] Edge cases considered (briefly note any)

- [ ] Before/after numbers included
- [ ] Benchmark command or link to CI perf output
- [ ] No unexpected regressions in CPU / memory / proof size / compile time

## Docs (if relevant)
- [ ] Rustdoc updated
- [ ] Book/README updated
- [ ] Examples updated

## Risk
Nothing, as this PR is only an RFC

## Review notes
None

### DIFF
--- a/docs/rfcs/0002-rfc-witness-removal.md
+++ b/docs/rfcs/0002-rfc-witness-removal.md
@@ -4,7 +4,7 @@
 
 - **Author(s):** @LindaGuiga
 - **DRI:** @LindaGuiga
-- **Status:** Draft
+- **Status:** Accepted
 - **Created:** 2026-01-13
 - **Tracking issue:** #103
 


docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
+# RFC 0002: Removing Witness Table and Supporting Base Field Elements
+
+---
+
+- **Author(s):** @LindaGuiga
+- **DRI:** @LindaGuiga
+- **Status:** Draft
+- **Created:** 2026-01-13
```
Nashtare: been waiting for some time I see 👀

docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
...
+- **Tracking issue:** #103
+
+## 1. Summary
+The Witness table is very long and could potentially blow up easily. Moreover, it only supports extension field elements, which is a problem for verifying uni-stark proofs.
```
Nashtare: w.r.t to the first sentence, for context, recursion gives with default packing a Witness table of length 2^19 after padding (around 420k/440k rows unpadded) with the second largest (MUL) is below 200k unpadded.

docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
...
+## 2. Motivation / Problem statement
+We currently have two issues with the design of the Witness table.
+
+The first is a performance issue related to the size of the Witness table. The latter could easily become very long and already takes the most time to prove. Moreover, its only goal is to ensure that values are not overwritten. While this is a critical soundness point, it could be ensured in other ways, as this RFC shows.
+
+The second is that the Witness table only supports extension field elements. On the other hand, the uni-stark challenger observes and retrieves base field elements. The current design therefore makes it impossible to correctly reproduce the behaviour of the uni-stark verifier. We could easily add 3 global lookups to the Witness table, but it would lead to at least 3 more columns for the multiplicities (and, for degree reasons, we would also have extra lookup polynomials). Since, as mentioned in the previous paragraph, the Witness table is currently very heavy, I do not think it would be advisable to implement this approach for extension field support. If we get rid of the Witness table, the impact of supporting extension field elements becomes lighter.
```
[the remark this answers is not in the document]
Nashtare: Instead I'd frame it as the current design (or rather the one in 244) while allowing it may make it harder / more complex than needed

docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
...
+### 4.1 High-level approach
+Let us first describe how we can get rid of the Witness table.
+
+The role of the Witness table is to ensure, through the CTLs, that all values associated to a certain index are equal. It corresponds to a lookup table into which all other tables read their inputs and outputs.
+
+To replace this pivotal role, I propose to keep the notion of witness indices and witness multiplicities, without storing them in any table. In each table, we can ensure that witness values are read-only as follows:
+- If we encounter an index for the first time (in any table), its multiplicity is positive.
+- If we have already encountered the index (in any table), then its multiplicity is negative.
+
+This way, all values pairs (index, value) are "reading" from the same location (the one with positive multiplicity).
```
Nashtare:
```suggestion
This way, all value pairs (index, value) are "reading" from the same location (the one with positive multiplicity).
```

docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
...
+
+The permutation table can get its inputs from the `Public` or `Constant` table. Since the challenger can observe base field values, `Public` and `Constant` either need to support both base field and extension field elements or only base field elements. The problem with the latter is that it would force `Add` and `Mul` to have 3 more CTLs per limb (and therefore 9 more permutation columns each). As `Public` and `Constant` should generally be shorter than `Add` and `Mul`, I believe it would be better for `Public` and `Const` to support both base field and extension field elements. `Add` and `Mul` can the remain unchanged -- with the exception of added multiplicities. I will explain later how this would work. 
+
+#### Permutation table
+For permutation inputs:
+- Since the base field input limbs are not necessarily contiguous, we need to have an index column for each limb. We would also need a CTL for each base field limb. Thus, we would need to have 12 extra index columns and 12 extra multiplicity colmns (note that we can use `in_ctl` to store the multiplicity of the first base field element).
```
Nashtare:
```suggestion
- Since the base field input limbs are not necessarily contiguous, we need to have an index column for each limb. We would also need a CTL for each base field limb. Thus, we would need to have 12 extra index columns and 12 extra multiplicity columns (note that we can use `in_ctl` to store the multiplicity of the first base field element).
```

docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
...
+
+### 4.2 APIs / traits / types
+The only trait / API changes are related to the way multiplicities are computed. When it comes to adapting the lookups to extension field elements, we only need to update `get_lookups` for all tables using what we described above. Of course, we would also need to update preprocessed values with the multiplicities.
+
+In order to keep track of the multiplicities for all tables, we can introduce a `WitnessMultiplicities` structure. Instead of a `Vec<F>`, `generate_preprocessed_columns` would update a `Vec<WitnessMultiplicities>`. In `WitnessMultiplicities>`, we need to keep track of the operation type and operation index where the index was created. Indeed, we need to update the multiplicity of that value every time the index is encountered again. Note that for the proposed approach to work, we need to ensure that indices are updated with increasing indices.
```
Nashtare:
```suggestion
In order to keep track of the multiplicities for all tables, we can introduce a `WitnessMultiplicities` structure. Instead of a `Vec<F>`, `generate_preprocessed_columns` would update a `Vec<WitnessMultiplicities>`. In `WitnessMultiplicities`, we need to keep track of the operation type and operation index where the index was created. Indeed, we need to update the multiplicity of that value every time the index is encountered again. Note that for the proposed approach to work, we need to ensure that indices are updated with increasing indices. 
```

docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
...
+pub struct Multiplicity {
+    pub extension: usize,
+    pub base: [usize; 4]
+}
+```
+which represents a multiplicity for extension and base field elements. The differentiation is only needed for `Public` and `Constant` though.
+
+We need to update `generate_preprocessed_columns` and `PrerocessedColumns`.
+
+We need to remove `Witness` from `PrimitieOpType`. The indexing needs to be modified as well (each base field element needs to get an index). In the following, we assume that indexing is correct. In fact, if we still assume that a `WitnessId` represents an extension field element, we can just assume that the witness index is a multiple of `D`, and the `Target` holds `D` base field elements.
```
Nashtare:
```suggestion
We need to remove `Witness` from `PrimitiveOpType`. The indexing needs to be modified as well (each base field element needs to get an index). In the following, we assume that indexing is correct. In fact, if we still assume that a `WitnessId` represents an extension field element, we can just assume that the witness index is a multiple of `D`, and the `Target` holds `D` base field elements.
```

docs/rfcs/0002-rfc-witness-removal.md
```diff
@@ -0,0 +1,258 @@
...
+                        base: [0; D]
+                    });
+                }
+                // Public: loads a public input into witness[out].
+                // Preprocessed data: the output witness index.
+                Op::Public { out, .. } => {
+                    let table_idx = PrimitiveOpType::Public as usize;
+
+                    // Since the values in `PublicAir` are looked up in `WitnessAir`,
+                    // we need to take the values into account in `WitnessAir`s preprocessed multiplicities.
```
Nashtare:
```suggestion
                    // we need to take the values into account in `WitnessAir`'s preprocessed multiplicities.
```

PR #297, opened 2026-02-21
feat: add a Horner accumulation step entry in `ALU` table

Nashtare:
## Summary

Add a `HornerAcc` entry in the ALU table, performing the accumulation step `result = acc * alpha + p_at_z - p_at_x`.
Because the Horner phase itself in `compute_single_reduced_opening()` represents a significant portion of the recursive verifier circuit, it makes sense offloading it to a custom ALU instruction.

@LindaGuiga I'll keep this on hold as this would most likely conflict with what you're doing for FRI, but opening it for visibility / in case this one gets favored.

### DIFF
--- a/book/src/appendix/benchmark.md
+++ b/book/src/appendix/benchmark.md
@@ -46,18 +46,18 @@ Running on a Apple M4 pro, 14 Cores, with **KoalaBear** field and extension of *
 
 - **Keccak AIR program:** (1,000 hashes)
   - Base uni-stark proof: 1.44 s
-  - 1st recursion layer: 1.29 s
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -66,11 +66,11 @@ jobs:
           cargo run --package p3-circuit-prover --example poseidon2_perm_chain
           cargo run --package p3-circuit-prover --example poseidon2_perm_merkle
 
-      - name: Run allocation_log example with debug logging
```
Nashtare: unrelated but needed to "actually" test the behavior now that this is behind a feature flag

PR #312, opened 2026-02-24
refactor: Implement RFC-0002 witness table removal

### LINKED ISSUE #103: Optimize Witness Table

Nashtare:
## Summary

Implement RFC 0002: remove the global Witness table and replace it with per-table multiplicities for Const/Public/ALU and Poseidon2 preprocessed data.

closes #103

## Why
The global Witness table was the largest  and slowest table and only enforced read-once consistency.

### DIFF
--- a/circuit-prover/benches/prover.rs
+++ b/circuit-prover/benches/prover.rs
@@ -4,18 +4,18 @@
 //! scaling is visible.
 
 use criterion::{BenchmarkId, Criterion, black_box, criterion_group, criterion_main};
-use p3_baby_bear::BabyBear;
 use p3_batch_stark::ProverData;
 use p3_circuit::CircuitBuilder;
 use p3_circuit_prover::air::{AluAir, ConstAir, PublicAir};
 use p3_circuit_prover::common::get_airs_and_degrees_with_prep;
-use p3_circuit_prover::config::BabyBearConfig;
+use p3_circuit_prover::config::KoalaBearConfig;
 use p3_circuit_prover::{
     BatchStarkProver, CircuitProverData, ConstraintProfile, TablePacking, config,
 };
 use p3_field::PrimeCharacteristicRing;
+use p3_koala_bear::KoalaBear;
 
-type F = BabyBear;
+type F = KoalaBear;
 
 fn fib_circuit(n: usize) -> (p3_circuit::Circuit<F>, F) {
     let mut builder = CircuitBuilder::new();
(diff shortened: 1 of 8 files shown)


circuit-prover/src/air/utils.rs
```diff
@@ -33,14 +33,53 @@ pub fn get_index_lookups<AB: PermutationAirBuilder + AirBuilderWithPublicValues,
         .collect()
 }
 
+/// D per-coefficient lookups on the "BaseFieldChecks" bus: `lookup_i: (index + i, coeff_i)` with multiplicity from preprocessed.
+pub fn get_base_field_lookups<AB, const D: usize>(
```
LindaGuiga: It doesn't seem to be used here (PublicAir doesn't look like it has base field lookups yet). So maybe we should put this in a follow-up PR?

circuit-prover/src/common.rs
```diff
@@ -173,66 +145,99 @@ where
...
             }
         }
     }
-    if mmcs_lookup_count > 0 {
-        tracing::debug!(
-            "Updated {} mmcs_index_sum lookups in witness multiplicities",
-            mmcs_lookup_count
-        );
+
+    // Phase 2: Update Poseidon2 out_ctl values in the base field preprocessed data.
```
LindaGuiga: Same as above + can't we do this in the same pass as the previous loop? (In order to update the correct creating table/index every time, we can also keep track of that in `ext_reads`...)

LindaGuiga (commented):
Looks good, thank you! I just have a few comments

circuit-prover/src/common.rs
```diff
@@ -164,7 +135,8 @@ where
     const MMCS_MERKLE_FLAG_OFFSET: usize = 21;
     const NEW_START_OFFSET: usize = 22;
 
-    let mut mmcs_lookup_count = 0usize;
+    // Phase 1: Scan Poseidon2 preprocessed data to count mmcs_index_sum conditional reads,
+    // and update `ext_reads` accordingly. This must happen before computing multiplicities.
     for (op_type, prep) in preprocessed.non_primitive.iter() {
```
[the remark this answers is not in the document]
Nashtare: The `mmcs_index_sum` lookup has a conditional multiplicity, which conflicts with Poseidon2 executor's logic when calling `preprocess()` for a single op, as it doesn't know what the next Poseidon2 op's `new_start` will be.

If we were to defer the count to a later pass inside `generate_preprocessed_columns()` we'd have lso issues depending on whether there was trace padding or not. The logic of padding from TablePacking (`min_trace_height` does not exist in the `p3-circuit` crate).

PR #347, opened 2026-03-09
perf: add a `Recompose` NPO for [BF; D] -> EF packing / unpacking

Nashtare:
## Summary

Define a Recompose NPO that handles the recomposition of extension fields from an array of base fields.
Circuits not defining this NPO at build time will fall back to the current method (sequence of MUL + ADD with the extension field generators via the ALU table).

The Recompose AIR does nothing (i.e. has no constraints). Correctness is enforced via lookups against the witness bus values. It's width is 4 main columns + 2 preprocessed ones.

## Why

Most of the current overhead lies in decomposing / recomposing extension field elements. See below for an overview of each scope of the recursive verifier circuit (as of commit 09a0b11):

Screenshot 2026-03-09 at 10 42 46

## Changes
- [x] Add a dedicated Recompose NPO instruction

## Tests
- [x] Unit tests added/updated
- [x] Existing tests pass

- [x] Updated performance benchmarks for `recursive_aggregation` test. Note that a new table comes with tradeoffs, and for small circuits like simple recursive chains (Fibonacci / Keccak), it seems performance decrease slightly with the new table, hence it is kept disabled there.

### DIFF
--- a/circuit-prover/src/batch_stark_prover/recompose.rs
+++ b/circuit-prover/src/batch_stark_prover/recompose.rs
@@ -66,15 +66,14 @@ impl<const D: usize> RecomposeProver<D> {
 
         // Build preprocessed data: 2 values per operation [output_idx, out_mult]
         // These are stored flat; create_direct_preprocessed_trace handles lane layout.
-        let mut preprocessed = Vec::with_capacity(num_ops * 2);
-        for row in &t.operations {
-            preprocessed.push(Val::<SC>::from_u32(row.output_wid.0 * D as u32));
-            preprocessed.push(Val::<SC>::ONE); // out_mult placeholder
+        let mut preprocessed = Val::<SC>::zero_vec(num_ops * 2);
+        for (i, row) in t.operations.iter().enumerate() {
+            preprocessed[i * 2] = Val::<SC>::from_u32(row.output_wid.0 * D as u32);
         }
 
         let air =
             RecomposeAir::<Val<SC>, D>::new_with_preprocessed(lanes, preprocessed, min_height);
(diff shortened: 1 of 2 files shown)


circuit-prover/src/batch_stark_prover/recompose.rs
```diff
@@ -0,0 +1,283 @@
...
+            preprocessed.downcast_mut::<PreprocessedColumns<BinomialExtensionField<F, 2>>>()
+        {
+            return recompose_preprocess_impl::<F, _>(prep);
+        }
+        if let Some(prep) = preprocessed.downcast_mut::<PreprocessedColumns<F>>() {
+            return recompose_preprocess_impl::<F, _>(prep);
+        }
+        Ok(HashMap::new())
+    }
+}
```
tcoratger: I guess that we could have something generic for this but let's do it in a followup PR :)

circuit-prover/src/batch_stark_prover/recompose.rs
```diff
@@ -0,0 +1,283 @@
...
+        let num_ops = t.total_rows();
+        let min_height = packing.min_trace_height();
+        let lanes = 1;
+
+        // Build preprocessed data: 2 values per operation [output_idx, out_mult]
+        // These are stored flat; create_direct_preprocessed_trace handles lane layout.
+        let mut preprocessed = Vec::with_capacity(num_ops * 2);
+        for row in &t.operations {
+            preprocessed.push(Val::<SC>::from_u32(row.output_wid.0 * D as u32));
+            preprocessed.push(Val::<SC>::ONE); // out_mult placeholder
```
tcoratger: Is it normal that we just have 1 as a placeholder here?

PR #374, opened 2026-03-12
circuit: add symbolic module with tests

tcoratger:
## Summary
What does this PR change?

@Nashtare More or less I think that in the mid/long-term future this new module could be upstreamed to Plonky3 (just not sure if we want to upstream the circuit builder as well in Plonky3 because it is somehow needed here).

Fixes #<issue-id>

## Why
What problem does this solve? Why is this approach correct?

## Changes
- [ ] Key change 1
- [ ] Key change 2

## Acceptance criteria
- [ ] <explicit condition that proves "done">
- [ ] <explicit condition that proves "done">

## Tests
- [ ] Unit tests added/updated
- [ ] Existing tests pass
- [ ] Edge cases considered (briefly note any)

- [ ] Before/after numbers included
- [ ] Benchmark command or link to CI perf output
- [ ] No unexpected regressions in CPU / memory / proof size / compile time

## Docs (if relevant)
- [ ] Rustdoc updated
- [ ] Book/README updated
- [ ] Examples updated

## Risk
What could break?

## Review notes
Where should reviewers focus on? Any tricky parts / follow-ups?


Nashtare (approved):
The refactoring is really neat thanks!

As for whether we would move parts of this codebase to upstream P3 in the mid/long term I'm not sure. This is already something that we were discussing with Angus and the team at the start of the project, the main reason against being the complexity and size of both independent projects already. 
This could of course be revisited in the future but I think I'd lean towards keeping the whole circuit building logic outside of Plonky3 as an independent workspace.

PR #439, opened 2026-05-05
perf: group matrices and reverse-horner chain

Nashtare:
Removes 3% of ALU ops in recursive aggregation.


LindaGuiga (commented):
LGTM, thank you!

PR #452, opened 2026-06-26
fix(circuit/lowerer): don't alias connect(value, const) into the constant's witness slot

emberian:
## What this fixes

In the circuit lowerer, declared `connect(a, b)` pairs are unioned in a DSU
(`ConnectDsu`) so that connected expressions share a single witness slot. That
aliasing is sound when both sides are value-bearing — they compute the same
value, so sharing a slot is free equality. It is **not** sound when exactly one
side is a constant. The dominant case is `assert_zero(x)`, which lowers to
`connect(x, ExprId::ZERO)`.

Aliasing `x` into the constant's class gives the value-bearing witness the
*constant's* shared slot. Because every `assert_zero` in a circuit routes through
the single circuit-wide `ExprId::ZERO` class, many independent `assert_zero`s
collapse into one shared slot. If any member is genuinely nonzero at
witness-generation time, its creator overwrites the shared constant slot and the
prover raises `WitnessConflict { WitnessId(0) }` — a spurious, hard-to-localize
failure rather than a clean, local mismatch. (We hit this as a ~219-member
`ZERO`-class collapse in a larger recursive circuit.)

## The fix

`connect` pairs with **exactly one** constant side are no longer fed to the DSU.
They are deferred and, after `emit_operations`, re-emitted as a per-target
equality constraint:

```text
Op::MulAdd { a: x, b: ZERO, c: const, out: x }   // x*0 + c == x  ⇔  x == c
```

This keeps `x` in its own witness slot, keeps the constant the sole writer of its
slot, and makes a mismatch fail locally on `x`'s slot. We chose `MulAdd` over a
second `Op::Const` because `ConstAir` treats every `Op::Const` as a slot creator
(which would reintroduce a double-creator), whereas `MulAdd` flows through the
existing ALU bus accounting with no new bus path. The equality is a genuine AIR
constraint, so it is *enforced*, not merely witness-aliased.

## Scope

Four files:

- `circuit/src/builder/compiler/lowerer/state.rs` — partition declared connects,
  defer value↔const pairs, add `emit_deferred_const_connects`.
- `circuit/src/builder/compiler/lowerer/mod.rs` — invoke it after
  `emit_operations`, before `backfill_connect_mappings`.
- `circuit/src/builder/compiler/lowerer/tests.rs` and
  `circuit/src/tables/runner.rs` — expected-output tests updated to the equality
  semantics (an `assert_zero` now produces one extra `MulAdd` equality row and
  keeps its target in its own slot).

## Testing

- `cargo test -p p3-circuit` — **323 passed, 0 failed**, off current `main`
  (`b81dc0d`) plus this commit.

## Honest framing (please read)

This was developed with substantial AI assistance. It passes the `p3-circuit`
suite but has **not** been independently audited by a STARK/recursion expert, and
it touches soundness-relevant accounting (witness-slot creation). We believe it
is correct and our tests agree, but please treat it as an unaudited proposal and
**review carefully**. If the framing is wrong or it isn't useful, just close — no
hard feelings.

We deliberately kept this small and easy to evaluate.

## Related work we'd be happy to upstream separately

While building an IVC light client on top of `Plonky3-recursion`, we also built a
couple of more general pieces on our fork that upstream currently lacks:

- A **verified public-output channel** for recursive proofs
  (`expose_as_public_output`): a bus-reader AIR that binds arbitrary in-circuit
  witnesses to host-readable public outputs of a recursive proof, provably equal
  to the genuine verified witnesses (so a light client can read accumulator
  state / chain heads / roots out of an aggregate proof).
- An **in-circuit verifier-key identity pin** for IVC fixed-points
  (`pin_preprocessed_commit`): constrains a child proof's preprocessed
  commitment to expected constants, so a from-scratch prover can't fold a proof
  of a *different* circuit — the stable anchor that makes "one running circuit
  verifies its own previous proof" close into a genuine fixed point.

Both are more general than our application, but they're larger and also touch
soundness, so we left them out of this PR to keep it focused. If either is of
interest we're glad to prepare separate PRs (with the same "unaudited, please
review" framing). No expectation either way.

### DIFF
--- a/circuit/src/builder/compiler/lowerer/mod.rs
+++ b/circuit/src/builder/compiler/lowerer/mod.rs
@@ -76,6 +76,12 @@ impl<'a, F: Field> ExpressionLowerer<'a, F> {
         state.emit_operations()?;
         // Verify that no registered non-primitive operation was left unreachable.
         state.validate_all_npo_emitted()?;
+        // Emit per-target equality constraints for the deferred `connect(value, const)` pairs
+        // (the `assert_zero(x)` family): these are NOT aliased into the constant's class, so the
+        // value keeps its own slot and a mismatch fails locally instead of clobbering the shared
+        // constant slot (the `WitnessId(0)` collapse). Must run AFTER `emit_operations` (every
+        // value/const target now has a witness slot) and BEFORE `backfill_connect_mappings`.
+        state.emit_deferred_const_connects()?;
         // Fill in witness mappings for connect-class members not directly visited.
         state.backfill_connect_mappings();
         // Convert the accumulated mutable state into the immutable result.
--- a/circuit/src/builder/compiler/lowerer/state.rs
+++ b/circuit/src/builder/compiler/lowerer/state.rs
@@ -50,6 +50,16 @@ pub(super) struct LoweringState<'a, F: Field> {
     /// Tracks which non-primitive operations have already been emitted,
     /// preventing duplicate emission from multiple output nodes.
     emitted_npo_ops: HashSet<NonPrimitiveOpId>,
+
+    /// `connect` pairs in which EXACTLY ONE side is a constant — normalized as
+    /// `(value_expr, const_expr)`. These are NOT fed to the DSU: aliasing a
+    /// value-bearing witness into a constant's shared slot is the unsound
+    /// representation behind the `WitnessId(0)` `assert_zero` collapse (a
+    /// genuinely-nonzero value-bearing witness gets the constant's slot and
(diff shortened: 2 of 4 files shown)


Nashtare:
I don't think this is a right way to approach the problem, this would make the circuit heavier for a problem that isn't one. Have you tried using the `debugging` feature and extracting the allocation logs (`builder.dump_allocation_log()`)? This may help your debugging without having to go through this.

emberian:
@Nashtare (thanks for the advice)

PR #453, opened 2026-06-26
feat(circuit): expose_as_public_output — a verified public-output channel for recursive proofs

emberian:
## What this adds

An opt-in non-primitive table, `expose_claim`, plus a builder method
`CircuitBuilder::expose_as_public_output(targets: &[ExprId])`, that surfaces
chosen in-circuit witnesses as **host-readable table public values, bound to the
genuine witnesses** via the `WitnessChecks` bus.

The table is a pure bus reader. For each chosen witness it receives the full
`[witness_idx, v_0, .., v_{D-1}]` tuple off the `WitnessChecks` bus with reader
multiplicity `-1` (so the bus stays balanced against the table that *created*
that witness), and constrains the table's public value for that lane to equal the
value it read. The host then reads `proof.non_primitives[].public_values` and is
guaranteed they are the genuine verified witnesses — not free prover-chosen
scalars.

## Why it's a general capability

As far as I can tell, upstream has no mechanism to bind an arbitrary in-circuit
witness to a host-readable *public output* of a recursive proof: `public_values`
/ `table_public_inputs` flow inward (values the AIR is checked against), and the
non-primitive tables (`poseidon2`, `recompose`, ...) all declare an empty
`public_values` vector.

A recursive/IVC light client that wants to read claims *out* of an aggregate
proof — a chain head, a Merkle root, a step counter, a running hash — needs
exactly this: an output channel whose values are provably the genuine in-circuit
witnesses. This is the abstracted primitive (the builder API + the AIR/op); the
specific orchestration we layer on top of it stays in our fork.

## On circuit weight (re: the discussion on #452)

This is **purely additive and opt-in**: the table only exists when a circuit
calls `enable_expose_claim` + `expose_as_public_output`. It adds **zero columns
and zero constraints** to any circuit that does not use it, and its own footprint
scales only with the number of values you choose to expose. It does not touch any
existing table. So it should not make anyone's circuit heavier — it is a new
capability you opt into, not a change to the default path. (One incidental fix:
`CircuitTableAir::num_public_values` now forwards to the dynamic AIR instead of
returning the default `0`; this is a no-op for every existing table, since they
declare no public values.)

## How it's tested

`batch_stark_prover::tests::test_babybear_expose_as_public_output`: a base-field
circuit exposes an input plus two internal computed witnesses (`a + b`, `a * b`),
and after `prove_all_tables` + `verify_all_tables` the host reads the three values
back out of the proof, asserted equal to the genuine in-circuit values. The full
`p3-circuit` (323) and `p3-circuit-prover` (46) lib suites pass off `main`.

## Caveat — please review carefully

This was developed with AI assistance while building an IVC layer on our fork. It
passes our suite but has **not been independently audited**, and it touches
soundness-relevant accounting (witness-slot reads, `WitnessChecks` bus
multiplicities, the public-value binding constraint). Please review with that in
mind. **Entirely happy for you to close this if it isn't a direction you want —
no hard feelings**; we can carry it on our fork either way.

This is a companion to #452 and part of the IVC-enablement work we built on our
fork. There is one further general piece we could offer as a follow-up if this is
useful: an in-circuit **VK-identity pin** (`pin_preprocessed_commit`) that binds a
child proof's preprocessed-commitment cap to expected constants, so a fold rejects
a proof of a *different* circuit — the stable anchor an IVC fixed-point needs. I
left it out here to keep this PR focused on one capability.

### DIFF
--- a/circuit-prover/src/air/expose_claim_air.rs
+++ b/circuit-prover/src/air/expose_claim_air.rs
@@ -1,0 +1,210 @@
+//! [`ExposeClaimAir`] surfaces N chosen witness values as table public values,
+//! BOUND to the genuine witnesses via the `WitnessChecks` bus.
+//!
+//! This is a verified public-output channel for a recursive circuit: given any
+//! set of in-circuit witnesses (each one SENT on the `WitnessChecks` bus by the
+//! table that created it — a primitive `PublicAir`, an ALU output, a Poseidon2
+//! limb, etc.), this table READS each of them back (reader multiplicity `-1`)
+//! into a main column and constrains the table's public value for that lane to
+//! equal the value it read. Because the read is bus-bound to the genuine witness,
+//! and the public value is locally constrained to the read, the host-readable
+//! `non_primitives[].public_values` are provably the genuine in-circuit values —
+//! not free prover-chosen scalars.
+//!
+//! A common use is to surface an aggregate's running state (a chain head, a
+//! Merkle root, a step counter, a running hash) out of a fold so a light client
+//! can read it; but the channel is generic over which witnesses are exposed.
+//!
+//! # Column layout (per lane = per exposed value)
+//!
+//! **Main columns** (D per lane): `v_0, ..., v_{D-1}` — the value read off the bus.
+//!
+//! **Preprocessed columns** per lane: `witness_idx`, `read_mult` (2 columns).
+//!
+//! **Public values**: one base-field public value per lane, in lane order.
+//!
+//! # Constraints (per lane)
+//!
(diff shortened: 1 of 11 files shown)


Nashtare:
@emberian did you mean to close it? I haven't looked at it yet but it could be a meaningful addition to the library.

emberian:
@Nashtare ah I had closed it maybe prematurely to author the "experience report" issue #454, feel free to reopen & we can iterate on it (there are other changes in the fork too: https://github.com/emberian/plonky3-recursion/tree/update-plonky3-rev#what-we-built-on-top-and-why-we-needed-it)

PR #455, opened 2026-07-04
fix: forward num_public_values in CircuitTableAir BaseAir impl

Raphsolt:
### What this does
Adds the missing `num_public_values` forwarding method to the `BaseAir` implementation for `CircuitTableAir`.

### Why it's needed
Currently, `CircuitTableAir` explicitly forwards `width`, `preprocessed_width`, and `preprocessed_trace`, but omits `num_public_values`. 

Because `BaseAir` provides a default implementation of `fn num_public_values(&self) -> usize { 0 }`, the compiler doesn't catch this omission. 

If a user defines a custom Non-Primitive Operation (NPO) table using the plugin system that *does* expose public values, `from_airs_and_degrees` incorrectly reads `0` for that table. This results in the symbolic builder receiving an empty public values slice, which causes a panic (`index out of bounds`) when the custom AIR attempts to evaluate constraints against its public values. 

This one-line match arm ensures custom dynamic AIRs correctly report their public value count to the prover.

### DIFF
--- a/circuit-prover/src/batch_stark_prover.rs
+++ b/circuit-prover/src/batch_stark_prover.rs
@@ -882,6 +882,7 @@ where
             }
         }
     }
+}
 
 macro_rules! impl_circuit_table_air_for_builder {
     ($builder_ty:ty) => {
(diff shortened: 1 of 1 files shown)


Nashtare (approved):
Thank you!

